PR #21, opened 2016-05-13
Clean-up + conf to properties

janvanmansum:


pom.xml
```diff
@@ -116,6 +120,10 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.jfrog.buildinfo</groupId>
```
rvanheest, line 132: What is this plugin all about?

Issue #15, opened 2016-03-14
Verify instructions.csv in the preconditions


rvanheest:
To minimize failure in the `Action.run()` calls, we should perform some validation on the `instructions.csv` file during the `Action.checkPreconditions` phase. This way we fail faster.

Things to validate:
* files that are mentioned in the `instructions.csv` file should exist in the `multi-deposit-directory`
* ...

[feel free to add stuff in the listing above]

lindareijnhoudt:
and vice versa? so files in the `multi-deposit-dir` should also be mentioned in the csv file?

rvanheest:
not sure about that one. I'd say probably yes, but not entirely sure about this. After all, it doesn't hurt the application if some files are in the multideposit directory and then not being mentioned in the `instructions.csv` file or used in the process of converting to a bag or sending to Springfield.

Maybe we should discuss this specific thing a bit more? Is there something stated about this in the specifications? Keep in mind that every validation takes time! We should not validate for things that are not relevant. This is especially true with large multideposits...

janvanmansum:
Not all files need to be specified in the `MDI` (Multi-Deposit Instructions) file, that is the reason for the default file handling. The files about which nothing is specified are processed according to the default file handling rules, for individual files the default can be overriden.

rvanheest:
Note that this is something that is not yet implemented! This is part of the [non-default file handling](https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/master/multi-deposit-instructions.md#non-default-processing), right?

lindareijnhoudt:
EASY-1015

PR #25, opened 2016-05-30
Non exhaustive pattern match

rvanheest:
a remainder of @jo-pol's pull request...

### DIFF
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
@@ -169,7 +169,7 @@ object AddDatasetMetadataToDeposit {
         case (None,    None,   Some(t)) => elem(s"dc:relation")(t)
         case _                          =>
           // TODO this case needs to be checked to not occur in the preconditions
-          // (see also comment in https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/master/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala#L32)
+          // (see also comment in https://github.com/DANS-KNAW/easy-split-multi-deposit/commit/dbda6cc2b78f93196be62b323a988e3781cb6926#diff-efd2dc8d9655ba9c6b577f13dd66627bR32)
           throw new IllegalArgumentException("preconditions should have reported this as an error")
       }
     }
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
@@ -167,7 +167,9 @@ object AddDatasetMetadataToDeposit {
         case (Some(q), None,   Some(t)) => elem(s"dcterms:$q")(t)
         case (None,    Some(l),_      ) => elem(s"dc:relation")(l)
         case (None,    None,   Some(t)) => elem(s"dc:relation")(t)
-        case _                          => // TODO do nothing; this needs to be checked in the preconditions
+        case _                          =>
+          // TODO this case needs to be checked to not occur in the preconditions (see also line 32)
+          throw new IllegalArgumentException("preconditions should have reported this as an error")
       }
     }
   }
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
@@ -167,6 +167,7 @@ object AddDatasetMetadataToDeposit {
         case (Some(q), None,   Some(t)) => elem(s"dcterms:$q")(t)
         case (None,    Some(l),_      ) => elem(s"dc:relation")(l)
         case (None,    None,   Some(t)) => elem(s"dc:relation")(t)
+        case _                          => // TODO do nothing; this needs to be checked in the preconditions
       }
     }
   }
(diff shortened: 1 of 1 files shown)


src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -167,6 +167,7 @@ object AddDatasetMetadataToDeposit {
         case (Some(q), None,   Some(t)) => elem(s"dcterms:$q")(t)
         case (None,    Some(l),_      ) => elem(s"dc:relation")(l)
         case (None,    None,   Some(t)) => elem(s"dc:relation")(t)
+        case _                          => // TODO do nothing; this needs to be checked in the preconditions
```
janvanmansum: @rvanheest @jo-pol: Does it make sense to throw an exception here, i.e. is it an error if execution ends up here anyway?

rvanheest: As the comment suggests, this is something that needs to be checked in the preconditions, so should not be happening _here_. That's why I choose to keep it empty here and leave a TODO for future @jo-pol or future me to implement that precondition together with multiple others. The main point of this PR is to fix a non-exhaustive pattern match that would cause errors if bad input was given...

janvanmansum: It is certainly good to fix the build warning, but I am not really in favour of keeping TODO's around in the code. They tend not to get fixed. If it is not too complex let's just find out what is appropriate to do here:
1. Throw an `IllegalStateException` (if we should never end up here)
2. Do nothing (if it is possible to legitimately end up here)

jo-pol: rephrase as 'preconditions should have reported this as an error', the real TODO is at line 32 and was mentioned in #15 (closed because its now in Jira)

janvanmansum: I don't see any rephrased comment. Also, again, if it is an error to get to that line, why not simply throw an exception?

rvanheest: I suggest the following: we throw an `IllegalArgumentException` here as this case indeed should never occur. However, I prefer to keep this TODO in the code for now as a reminder to the one who is going to fix issue #15. Let's also add a notice about this TODO in the Jira issue.

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -167,6 +167,9 @@ object AddDatasetMetadataToDeposit {
         case (Some(q), None,   Some(t)) => elem(s"dcterms:$q")(t)
         case (None,    Some(l),_      ) => elem(s"dc:relation")(l)
         case (None,    None,   Some(t)) => elem(s"dc:relation")(t)
+        case _                          =>
+          // TODO this case needs to be checked to not occur in the preconditions (see also line 32)
```
janvanmansum: @rvanheest : "see also line 32". This is bound to become a "lying comment" very soon!

rvanheest: I know, but there is not really a way to refer to comments in another way, right? If you got a better way to refer to line 32, I'll be happy to change it!

jo-pol: It requires a bit of browsing through github (history an view button) or figuring out a commit SHA otherwise and assemble the link manually. You don't need the full SHA and can specify a range of lines with #L31-#l33 not sure a commit SHA might get lost with a squash merge and dropped branches, but github might preserve commits of a pull request even when its branch is dropped.

https://github.com/rvanheest/easy-split-multi-deposit/blob/6f17388233360f47f1616f42e458f5fee461987b/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala#L32

rvanheest: Gave the solution by @jo-pol some thought and I think this is the best solution for now. Instead of the link inside this PR, I used the link to master, which already contains this TODO comment.

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -168,7 +168,8 @@ object AddDatasetMetadataToDeposit {
         case (None,    Some(l),_      ) => elem(s"dc:relation")(l)
         case (None,    None,   Some(t)) => elem(s"dc:relation")(t)
         case _                          =>
-          // TODO this case needs to be checked to not occur in the preconditions (see also line 32)
+          // TODO this case needs to be checked to not occur in the preconditions
+          // (see also comment in https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/master/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala#L32)
```
jo-pol: referting to master is as volatile as it used to be, you _do_ need a commit sha

rvanheest: Fine with me as well, but you could also delete that commit by altering the history... And also, when you refer to a commit in a branch and you do a PR-squash-merge, I guess that commit doesn't exist anymore as well, right?

I'll add a commit SHA of the latest commit on master, expecting that this commit shouldn't vanish soon.

jo-pol: Not sure but I guess github preserves pull request, even if you drop the branch, take a look at previously squash-merged pull request. Anyhow, linking to the master branch just makes the comment longer and doesn't add any persistence.

rvanheest: See the latest commit: a4a8b08968919135b50006b0c466533ea9b956cf?

PR #29, opened 2016-06-28
incomplete enumeration of mandatory columns

jo-pol:
fixes incomplete documentation
#### When applied it will
- tell the user all DDM columns are required

trivial

Try a split-multi-deposit with or without one of de ddm columns followed with an easy-ingest-flow.

| repo | PR |
| --- | --- |
| easy- | [PR#](PRlink) |

### DIFF
--- a/multi-deposit-instructions.md
+++ b/multi-deposit-instructions.md
@@ -68,7 +68,7 @@ The supported metadata elements are subdivided into the following groups:
 
 The following elements are required: `DC_TITLE`, `DC_DESCRIPTION`, `DCX_CREATOR_*`
 (at least both the subfields `DCX_CREATOR_INITIALS` and `DCX_CREATOR_SURNAME` *or*
-the subfield `DCX_CREATOR_ORGANIZATION`), `DDM_*`.
+the subfield `DCX_CREATOR_ORGANIZATION`), `DDM_CREATED`, `DDM_AUDIENCE`, `DDM_ACCESSRIGHTS`.
 
 ### Semantics of the columns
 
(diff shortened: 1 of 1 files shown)


janvanmansum:
@jo-pol: Just to be sure: have you verified that all the DDM-fields are indeed mandatory?

jo-pol:
I discovered the omission because I I tried a minimal test. We don't have many: `DDM_CREATED`, `DDM_AVAILABLE`, `DDM_AUDIENCE`, `DDM_ACCESSRIGHTS`, `DDM_AUDIENCE`, `DDM_ACCESSRIGHTS`

lindareijnhoudt:
there are 6 fields mandatory in the DDM. 
- description
- title
- creator
- DDM_CREATED, 
- DDM_AUDIENCE, 
- DDM_ACCESSRIGHTS, 

the csv file does not prefix them all with DDM_ though, the title, description and creator are in DC_ and were already mentioned as required elements, so that's ok.

DDM_AVAILABLE is not mandatory, so we do need a list of DDM fields that are required and not a catch-all.

rvanheest:
Is this README change reflected by the code. A.k.a. does SMD complain if you leave out mandatory fields? And is the error message appropriate?

jo-pol:
The error is thrown when ingesting, not by split-multi-deposit. Anyway it is something the user should know. Error handling/reporting is poor on more aspects.

rvanheest:
It is something the user should know, indeed! Preferably as early as possible! So maybe we should consider adding these checks here. Wasn't there an issue in Jira that addresses this?

jo-pol:
You mean [1015](https://drivenbydata.atlassian.net/browse/EASY-1015), it doesn't include these type of checks.

rvanheest:
Then I'd say we should fix it in this PR. We should keep the README consistent with the application's behavior.

PR #32, opened 2016-08-10
(EASY-1106) Add mimetypes for microsoft documents that are not in default java.ne…

lindareijnhoudt:
fixes EASY-1106
#### When applied it will
- Add mimetypes for microsoft documents that are not in default java.net FileNameMap

AddFileMetadataToDepositSpec.scala contains a test for a docx file.


lindareijnhoudt:
i don't understand why this test
"datasetToFileXml" should "produce the xml for all the files" 
still works, even though i added a file to the "reisverslag" directory ...
shouldn't it now show 4 files instead of 3?

janvanmansum:
@lindareijnhoudt The test you are talking about looks in "Roundtrip_MD". You added the file in "Roundtrip".

PR #37, opened 2016-10-11
Removed parts of Javadocs that caused build warnings

janvanmansum:
No JIRA-issue.

The build displayed a lot of warnings because of links in the Javadocs it couldn't create. These links were references to the Java JRE javadocs, Apache commons, etc. I chose the easiest solution: delete those references.

### DIFF
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
@@ -66,14 +66,14 @@ package object multideposit {
   implicit class StringExtensions(val s: String) extends AnyVal {
     /**
       * Checks whether the `String` is blank
-      * (according to [[org.apache.commons.lang.StringUtils.isBlank]])
+      * (according to org.apache.commons.lang.StringUtils.isBlank)
       *
       * @return
       */
     def isBlank = StringUtils.isBlank(s)
 
     /** Converts a `String` to an `Option[String]`. If the `String` is blank
-      * (according to [[org.apache.commons.lang.StringUtils.isBlank]])
+      * (according to org.apache.commons.lang.StringUtils.isBlank)
       * the empty `Option` is returned, otherwise the `String` is returned
       * wrapped in an `Option`.
       *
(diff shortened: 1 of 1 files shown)


janvanmansum:
@DANS-KNAW/easy for review.

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
```diff
@@ -124,7 +124,6 @@ package object multideposit {
       * Writes a CharSequence to a file creating the file if it does not exist using the default encoding for the VM.
       *
       * @param data the content to write to the file
-      * @throws IOException in case of an I/O error
```
rvanheest, line 127: I really like to have these `@throws` here to remind us all that these methods can throw. On the other hand, we could also choose to go the other route of wrapping it in `Try`, but that would require some more effort (maybe make a JIRA issue for that?).

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
```diff
@@ -191,16 +186,13 @@ package object multideposit {
...
       * it is not guaranteed that those operations will succeed.
       * If the modification operation fails, no indication is provided.
       *
       * @param destDir the new directory, must not be ``null``
-      * @throws NullPointerException if source or destination is ``null``
-      * @throws IOException if source or destination is invalid
-      * @throws IOException if an IO error occurs during copying
       */
-    def copyFile(destDir: File) = FileUtils.copyFile(file, destDir)
+    def copyFile(destDir: File): Unit = FileUtils.copyFile(file, destDir)
```
rvanheest: why add a return type here and not in all other functions?

rvanheest (changes requested):
Funny, @lindareijnhoudt and I just ran into this too, this morning... See the 2 comments of mine below.

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
lindareijnhoudt: i agree. it seems to be a scaladoc bug. i see two options: 
- add a -no-link-warnings to the scaladoc, 
- or remove the throws-clauses from the docs

@janvanmansum has opted for the second one. i've added throws annotations to keep the reference in the code

rvanheest: I think the addition of the annotations makes it kind of equivalent to the original. So definitely 👍 for that!

PR #38, opened 2016-10-17
Easy-1134: Split the path to the datafile correctly

lindareijnhoudt:
fixes EASY-1134
#### When applied it will
- Split the path to the datafile correctly, regardless of directory-names
- Add a unittest for the bug described in EASY-1134


rvanheest (approved):

janvanmansum (approved):
Some of the names in this project are rather confusing, e.g. `xmlPerPath` (`getFileItemxml`?) and `multiDepositDir` (`getDatasetDir`?).

Anyhow, that has no bearing on this issue.

PR #43, opened 2017-01-25
EASY-1189: Project refactoring

rvanheest:
fixes EASY-1189

@DANS-KNAW/easy please review

Consider this PR to be a rewrite of most of the program. Don't look at the diff, but checkout the code and look through the code in IntelliJ. I'll do a demo of the changes in this PR on Monday.

#### Most notable changes
* RxScala is removed from the project
* The logic of an Action is contained in the Action class itself
* The main is considerably more easy to read and understand

#### TODO
* [ ] AddDatasetMetadataToDeposit, check on `DCX_RELATION_XXX` fields and add it to the `checkPreconditions`
* [ ] *???* make a command in the tool that lists all valid headers *???*


jo-pol:
I guess the folder `src/main/assembly/dist/res` is obsolete.

rvanheest:
Given that these files are added (and never edited afterwards) during the initial commit of this repository, I expect they were just copied from another project. Given that they're never used, I'll delete them.

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -15,32 +15,155 @@
...
+  private[Action] def innerCheckPreconditions: Try[Unit] = Try(logPreconditions()).flatMap(_ => checkPreconditions)
+  private[Action] def innerExecute(): Try[Unit] = Try(logExecute()).flatMap(_ => execute())
+  private[Action] def innerRollback(): Try[Unit] = Try(logRollback()).flatMap(_ => rollback())
+
+  /**
+   * Verifies whether all preconditions are met for this specific action.
+   *
+   * @return `Success` when all preconditions are met, `Failure` otherwise
+   */
+  protected def checkPreconditions: Try[Unit] = Success(())
```
jo-pol, line 53: why not `checkPreconditions()` just like the other two?

rvanheest: just Scala convention: `checkPreconditions` is a _pure_ function that does not modify any data, do any I/O or anything like that (it has no side-effects). `execute` and `rollback` however are expected to be _unpure_ functions, because they are supposed to write data to files, create folders on the filesystem and undo all that stuff.

See also http://stackoverflow.com/questions/3877953/why-to-use-empty-parentheses-in-scala-if-we-can-just-use-no-parentheses-to-defin

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -15,32 +15,155 @@
...
-  def rollback(): Try[Unit] = Success(())
+   * Compose this action with `other` into a new action such that it runs both actions in the correct order.
+   * [[checkPreconditions]] and [[execute]] are run in order, [[rollback]] is run in reversed order.
+   *
+   * @param other the second action to be run
+   * @return a composed action
+   */
+  def compose(other: Action): ComposedAction = {
+    other match {
+      case composedAction: Action#ComposedAction => ComposedAction(this :: composedAction.actions)
```
jo-pol, line 129: You thought I used the `#` too, but I'm using a [dot](https://github.com/d-bl/GroundForge/blob/0f39fa2/src/main/scala/dibl/ThreadDiagram.scala#L31-L31), so I don't understand. 

PS: The difference I see is you are dealing with a final class in a trait, while I am dealing with a plain case class in an object.

rvanheest: Yup, in your project `Force` is an object, so it's an instance (singleton) as well. Here, however, `Action` is a trait rather than an object, so it's _not_ an instance. Then you use `#` instead.

See:
* http://stackoverflow.com/questions/6676048/why-does-one-select-scala-type-members-with-a-hash-instead-of-a-dot
* http://stackoverflow.com/questions/12959438/investigation-of-type-and-keywords-in-scala

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -15,32 +15,155 @@
...
+   */
+  final def run: Try[Unit] = {
+    for {
+      _ <- innerCheckPreconditions.recoverWith {
+        case NonFatal(e) =>
+          Failure(PreconditionsFailedException(generateReport("Precondition failures:", e, "Due to these errors in the preconditions, nothing was done.")))
+      }
+      _ <- innerExecute().recoverWith {
+        case NonFatal(e) => List(Failure(e), innerRollback()).collectResults.recoverWith {
+          case e: CompositeException => Failure(ActionRunFailedException(generateReport("Errors in Multi-Deposit Instructions file:", e)))
```
jo-pol: You got past the preconditions, so is it really an error in the instructions file? Could be an out of disk space or whatever.

Don't we need a footer about changes are rolled back or whatever is appropriate?

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -15,32 +15,155 @@
...
+        }
+      }
+    } yield ()
+  }
+
+  private def generateReport(header: String = "", t: Throwable, footer: String = ""): String = {
+    val report = t match {
+      case ActionException(row, msg, _) => s" - row $row: $msg"
+      case CompositeException(ths) =>
+        ths.toSeq
```
jo-pol: You seem to drop an `i` to avoid a conflict with the keyword `this`. Rather use a longer more meaningful name to help the reader. Ah, checking the type you mean the plural _throwable_**s**

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -15,32 +15,155 @@
...
-    */
-  def run(): Try[Unit]
+   * Run an action. First the precondition is checked. If it fails a `PreconditionsFailedException`
+   * with a report is returned. Else, if the precondition succeed, the action is executed.
+   * If this fails, the action is rolled back and a `ActionRunFailedException` with a report is returned.
+   * If the execution was successful, `Success` is returned
+   *
+   * @return `Success` if the full execution was successful, `Failure` otherwise
+   */
+  final def run: Try[Unit] = {
```
jo-pol, line 77: Getting lost in all the brackets
```
    for {
      _ <- innerCheckPreconditions.recoverWith {
        case NonFatal(e) =>
          Failure(PreconditionsFailedException(generateReport(
            header = "Precondition failures:",
            t = e,
            footer = "Due to these errors in the preconditions, nothing was done."
          )))
      }
      _ <- innerExecute().recoverWith {
        case NonFatal(e) =>
          List(Failure(e), innerRollback())
            .collectResults.recoverWith {
            case e: CompositeException =>
              Failure(ActionRunFailedException(generateReport(
                header = "Errors in Multi-Deposit Instructions file:",
                t = e
              )))
          }
      }
    } yield ()
```

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -15,32 +15,155 @@
...
+  }
+
+  private def generateReport(header: String = "", t: Throwable, footer: String = ""): String = {
+    val report = t match {
+      case ActionException(row, msg, _) => s" - row $row: $msg"
+      case CompositeException(ths) =>
+        ths.toSeq
+          .collect { case ActionException(row, msg, _) => s" - row $row: $msg" }
+          .mkString("\n")
+      case e => s" - unexpected error: ${e.getMessage}"
```
jo-pol: Has the stack trace been logged? For unexpected errors we should not loose it.

src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
```diff
@@ -75,17 +73,22 @@ class ScallopCommandLine(props: PropertiesConfiguration, args: Array[String]) ex
            |Options:
            |""".stripMargin)
 
-  val multiDepositDir = trailArg[File](name = "multi-deposit-dir", required = true,
+  val multiDepositDir: ScallopOption[File] = trailArg[File](
```
jo-pol, line 77: I had a change hanging around to make the help layout consistent with other components:

```
  banner(s"""
           |  Utility to process a Multi-Deposit prior to ingestion into the DANS EASY Archive
           |
           |Usage:
           |  $printedName.sh [{--springfield-inbox|-s} <dir>] <multi-deposit-dir> <output-deposits-dir>
           |
           |Options:
           |""".stripMargin)```

File: src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala [hunk 8]
```diff
@@ -56,7 +54,7 @@ object CommandLineOptions {
         LdapImpl(new InitialLdapContext(env, null))
       })
 
-    log.debug(s"Using the following settings: $settings")
+    debug(s"Using the following settings: $settings")
```
jo-pol, line 55: Not sure any more this doesn't log passwords. Though that might not be a problem when originating from `application.properties` anyway.

src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
```diff
@@ -19,24 +19,22 @@ import java.io.File
...
+    debug("Loading application.properties ...")
     val homeDir = new File(System.getProperty("app.home"))
     val props = {
       val ps = new PropertiesConfiguration()
       ps.setDelimiterParsingDisabled(true)
       ps.load(new File(homeDir, "cfg/application.properties"))
       ps
     }
-    log.debug("Parsing command line ...")
+    debug("Parsing command line ...")
```
jo-pol, line 37: Not your doing but what about changing
```
      ldap = {
        import java.{util => ju}

        val env = new java.util.Hashtable[String, String]
```
into
```
      ldap = {
        val env = new java.util.Hashtable[String, String]
```

src/main/scala/nl/knaw/dans/easy/multideposit/Main.scala
```diff
@@ -16,188 +16,89 @@
...
-    def runActions = Main.runActions(actions)
+    logger.debug("closing ldap")
+    settings.ldap.close()
   }
 
-  implicit class DatasetsExtensionMethods(val datasets: Observable[(DatasetID, Dataset)]) extends AnyVal {
-    def getActions(implicit s: Settings) = Main.getActions(datasets)
+  def run(implicit settings: Settings): Try[Unit] = {
+    MultiDepositParser.parse(multiDepositInstructionsFile)
+      .flatMap(getActions(_).reduce(_ compose _).run)
```
jo-pol, line 43: Is an empty instruction file an error and reported before?

src/main/scala/nl/knaw/dans/easy/multideposit/Main.scala
```diff
@@ -16,188 +16,89 @@
...
 
-    dss.map(getActionsPerEntry)
-      .foldLeft(Result())(addToResult)
-      .flatMap(addCreateSpringfieldActions) // because of foldLeft above, this function is only executed once
+  def getGeneralActions(datasets: Datasets)(implicit settings: Settings): Seq[Action] = {
+    Seq(CreateSpringfieldActions(-1, datasets))
   }
 
-  def getDatasetActions(entry: (DatasetID, Dataset))(implicit settings: Settings): Observable[Try[Action]] = {
+  def getDatasetActions(entry: (DatasetID, Dataset))(implicit settings: Settings): Seq[Action] = {
```
jo-pol: `val (datasetID, dataset) = entry` and below that occurrence with double-r

src/main/scala/nl/knaw/dans/easy/multideposit/Main.scala
```diff
@@ -16,188 +16,89 @@
...
-          val check = action.checkPreconditions
-
-          if (check.isFailure) fails :+ check else fails
-        })
-      }
-      def toObservable = {
-        if (fails.isEmpty) actions.toObservable
-        else Observable.error(new Exception(generateErrorReport("Precondition failures:", fails,
-          "Due to these errors in the preconditions, nothing was done.")))
+          if isThisAudioVideo matches "(?i)yes" => CopyToSpringfieldInbox(row, fileMd)
```
jo-pol: new line at `=>`?

src/main/scala/nl/knaw/dans/easy/multideposit/MultiDepositParser.scala
```diff
@@ -17,79 +17,77 @@ package nl.knaw.dans.easy.multideposit
...
     val springfieldHeaders = List("SF_DOMAIN", "SF_USER", "SF_COLLECTION", "SF_PRESENTATION", "SF_SUBTITLES")
     val administrativeHeaders = List("DEPOSITOR_ID")
 
     val validHeaders = "DATASET" :: fileHeaders ++ springfieldHeaders ++ administrativeHeaders ++ DDM.allFields
-    if (headers.forall(validHeaders.contains)) Success(Unit)
-    else Failure(new ActionException(0, "SIP Instructions file contains unknown headers: "
-      + headers.filter(!validHeaders.contains(_)).mkString(", ") + ". "
+    val invalidHeaders = headers.filterNot(validHeaders.contains)
+    if (invalidHeaders.isEmpty) Success(Unit)
+    else Failure(ActionException(0, s"SIP Instructions file contains unknown headers: ${invalidHeaders.mkString(", ")}. "
```
jo-pol, line 79: We have inconsistent behaviour between components, [one](https://github.com/DANS-KNAW/easy-update-metadata-dataset/blob/8e95874/src/main/scala/nl/knaw/dans/easy/umd/InputRecord.scala#L27-L50) requiring a specific order, the other not. Which one should change?

src/main/scala/nl/knaw/dans/easy/multideposit/MultiDepositParser.scala
```diff
@@ -17,79 +17,77 @@ package nl.knaw.dans.easy.multideposit
...
-    * @param headers the headers to be validated
-    * @return `Success` if the `headers` are valid, `Failure` if the `headers` are invalid
-    */
+   * Tests whether the given list of headers is valid. If so, `Success(Unit)` is returned,
+   * else a `Failure` with a detailed `ActionException` is returned.
+   *
+   * @param headers the headers to be validated
+   * @return `Success` if the `headers` are valid, `Failure` if the `headers` are invalid
+   */
+  // TODO valid headers listing as sub-command???
```
jo-pol, line 70: We have inconsistent behaviour between components, [one](https://github.com/DANS-KNAW/easy-update-metadata-dataset/blob/8e95874/src/main/scala/nl/knaw/dans/easy/umd/InputRecord.scala#L27-L50) requiring a specific order, the other not. Which one should change?

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
```diff
@@ -96,201 +98,216 @@ package object multideposit {
...
+
+    def ifFailure(f: PartialFunction[Throwable, Unit]): Try[T] = {
+      t match {
+        case failure@Failure(e) if f.isDefinedAt(e) => Try {
+          f(e)
+          return failure
+        }
+        case x => x
+      }
+    }
```
jo-pol, line 133: I'm lost with these. 

The first is only use in one source, the others just once in another, so why have them here?

Why not stick to the usual `map` and `recover`?

jo-pol (commented):
So far the sources in the main package.

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
rvanheest: looks better indeed. I also changed `t` to `throwable`; seemed a better description...

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
rvanheest: You're right. That was an old text. Changed it.

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
rvanheest: Good point! Now it does...
I added a `cause` to the `ActionRunFailedException` and the `PreconditionsFailedException` and fixed this in the `run`. If you run the logger in debug mode, the stack traces will show up in the log.

src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
rvanheest: I see what you mean. `$settings` calls the `toString`, which only prints `multidepositDir`, `springfieldInbox` and `outputDepositDir`. So, no passwords from LDAP are printed!

src/main/scala/nl/knaw/dans/easy/multideposit/Main.scala
rvanheest: it did, but only because the parser calls `xs.head`, which throws a `NoSuchElementException` if `xs` is empty. I changed that by replacing the header validation with:

```scala
content.headOption.map(validateDatasetHeaders).getOrElse(Failure(EmptyInstructionsFileException(file)))
```

src/main/scala/nl/knaw/dans/easy/multideposit/Main.scala
rvanheest: changed it!

src/main/scala/nl/knaw/dans/easy/multideposit/MultiDepositParser.scala
rvanheest: neither, IMO. The one you refer to maps to an object; this project maps its values to another kind of data structure (`HashMap`)

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
rvanheest: These are for side-effects specifically. I'll move these to the dans-scala-lib some day.

PR #45, opened 2017-02-09
Fix easy 1166

PaulBoon:
fixes EASY-1166

#### When applied it will
* using an extra trailing argument for the datamanager username, it will retrieve the email address 

AddPropertiesToDeposit

run the command with a datamanager that is an active user and an activist and then look into the property files in the bags

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,6 +135,11 @@
             <plugin>
                 <groupId>com.mycila</groupId>
                 <artifactId>license-maven-plugin</artifactId>
+                <configuration combine.self="override">
+                    <excludes>
+                        <exclude>data/</exclude>
+                    </excludes>
+                </configuration>
             </plugin>
             <plugin>
                 <artifactId>maven-release-plugin</artifactId>
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
@@ -21,7 +21,9 @@ import javax.naming.ldap.InitialLdapContext
 
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
-import org.rogach.scallop.{ ScallopConf, ScallopOption }
+import org.rogach.scallop.{ScallopConf, ScallopOption}
+
+import scala.util.{Failure, Success, Try}
 
 object CommandLineOptions extends DebugEnhancedLogging {
 
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
@@ -68,12 +68,15 @@ class ScallopCommandLine(props: PropertiesConfiguration, args: Array[String]) ex
 
   printedName = "easy-split-multi-deposit"
   version(s"$printedName ${Version()}")
+  val description = "Splits a Multi-Deposit into several deposit directories for subsequent ingest into the archive"
+  val synopsis = s"""$printedName.sh [{--springfield-inbox|-s} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>"""
   banner(s"""
-           |Splits a Multi-Deposit into several deposit directories for subsequent ingest into the archive
-           |Utility to process a Multi-Deposit prior to ingestion into the DANS EASY Archive
+           |  $description
+           |  Utility to process a Multi-Deposit prior to ingestion into the DANS EASY Archive
            |
            |Usage:
-           |  $printedName.sh [{--springfield-inbox|-s} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
+           |
+           |  $synopsis
            |
            |Options:
            |""".stripMargin)
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
@@ -41,6 +41,7 @@ object CommandLineOptions extends DebugEnhancedLogging {
       multidepositDir = opts.multiDepositDir(),
       springfieldInbox = opts.springfieldInbox(),
       outputDepositDir = opts.outputDepositDir(),
+      datamanager = opts.datamanager(),
       ldap = {
         val env = new java.util.Hashtable[String, String]
         env.put(Context.PROVIDER_URL, props.getString("auth.ldap.url"))
@@ -69,7 +70,7 @@ class ScallopCommandLine(props: PropertiesConfiguration, args: Array[String]) ex
            |Utility to process a Multi-Deposit prior to ingestion into the DANS EASY Archive
            |
            |Usage:
-           |  $printedName.sh [{--springfield-inbox|-s} <dir>] <multi-deposit-dir> <output-deposits-dir>
+           |  $printedName.sh [{--springfield-inbox|-s} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
            |
            |Options:
            |""".stripMargin)
@@ -92,6 +93,11 @@ class ScallopCommandLine(props: PropertiesConfiguration, args: Array[String]) ex
     required = true,
     descr = "A directory in which the deposit directories must be created. "
       + "The deposit directory layout is described in the easy-sword2 documentation")
+
+  val datamanager: ScallopOption[String] = trailArg[String](
+    name = "datamanager",
+    required = true,
+    descr = "The user id of the datamanger (archivist) performing this deposit")
 
   validateFileExists(multiDepositDir)
   validateFileIsDirectory(multiDepositDir)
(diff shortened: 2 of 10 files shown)


pom.xml
```diff
@@ -135,6 +135,11 @@
             <plugin>
                 <groupId>com.mycila</groupId>
                 <artifactId>license-maven-plugin</artifactId>
+                <configuration combine.self="override">
+                    <excludes>
+                        <exclude>data/</exclude>
```
rvanheest, line 140: @janvanmansum we should probably add this to the microservice projects and the archetype as well. @PaulBoon and I experienced problems with the license plugin when we had a `home/` and `data/` folder coming from `debug-reset-apphome.sh` whenever we ran `mvnci`

src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
```diff
@@ -93,6 +96,11 @@ class ScallopCommandLine(props: PropertiesConfiguration, args: Array[String]) ex
     descr = "A directory in which the deposit directories must be created. "
       + "The deposit directory layout is described in the easy-sword2 documentation")
 
+  val datamanager: ScallopOption[String] = trailArg[String](
+    name = "datamanager",
+    required = true,
+    descr = "The user id of the datamanger (archivist) performing this deposit")
```
rvanheest: 'user id' --> 'username' ???

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -66,17 +111,20 @@ object AddPropertiesToDeposit {
       .getOrElse(Failure(ActionException(row, """The column "DEPOSITOR_ID" is not present""")))
   }
 
-  def writeProperties(row: Int, datasetID: DatasetID, dataset: Dataset)(implicit settings: Settings): Try[Unit] = {
-    val props = new Properties
+  def writeProperties(row: Int, datasetID: DatasetID, dataset: Dataset, emailaddress: String)(implicit settings: Settings): Try[Unit] = {
+    val props = new Properties {
+      // Make sure we get sorted output, which is better readable than random
+      override def keys(): util.Enumeration[AnyRef] = Collections.enumeration(new util.TreeSet[Object](super.keySet()))
```
rvanheest: can't you just write these without `util`? If you explicitly want them to indicate that you are talking about a Java class, make it `ju.Enumeration` and `ju.TreeSet`, as is conventional. You do this by adding this line: `import java.{util => ju}`

I like that you create a sorted `Properties` output with this!

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -15,33 +15,78 @@
...
 
-  /**
-   * @inheritdoc
-   * Checks whether there is only one unique DEPOSITOR_ID set in the `Dataset` (there can be multiple values but the must all be equal!).
-   * @return `Success` when all preconditions are met, `Failure` otherwise
-   */
-  override def checkPreconditions: Try[Unit] = validateDepositor(row, datasetID, dataset)
+  override def checkPreconditions: Try[Unit] = {
+    List(validateDepositor(row, datasetID, dataset), getDatamanagerMailadres(row))
+        .collectResults.map(_=>())
```
rvanheest: formatting...

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -15,33 +15,78 @@
...
+  }
 
-  override def execute(): Try[Unit] = writeProperties(row, datasetID, dataset)
+  override def execute(): Try[Unit] = getDatamanagerMailadres(row).flatMap(email => writeProperties(row, datasetID, dataset, email))
```
rvanheest: `writeProperties(row, datasetID, dataset, _)`

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -15,33 +15,78 @@
...
+        .flatMap(attrsSeq => {
+          if (attrsSeq.isEmpty) Failure(new ActionException(row, s"""DatamanagerID "$id" is unknown"""))
+          else if (attrsSeq.size > 1) Failure(new ActionException(row, s"""There appear to be multiple users with id "$id""""))
+          else Success(attrsSeq.head)
+        })
+        .flatMap(attrs => {
+          Option(attrs.get("dansState"))
+            .filter(_.get.toString == "ACTIVE")
+            .map(_ => Success(attrs))
+            .getOrElse(Failure(new ActionException(row, s"""The datamanager "$id" is not an active user""")))
```
rvanheest, line 70: Now both the fact that `dansState` does not exist and the fact that the state is not equal to 'ACTIVE' will result in the same `Failure`. We discussed this, because you had an if-else here initially and you had convinced me that that was the better way. So I'm a bit surprised to see this here 😉 . Nevertheless I like this solution as well, so if you do not care about the two facts resulting in the same `Failure`, I'm OK with it!

Same for the two instances below

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -66,17 +111,20 @@ object AddPropertiesToDeposit {
...
-    addProperties(props, dataset)
+    addProperties(props, dataset, emailaddress)
       .flatMap(_ => Using.fileWriter(encoding)(outputPropertiesFile(datasetID)).map(out => props.store(out, "")).tried)
       .recoverWith {
         case NonFatal(e) => Failure(ActionException(row, s"Could not write properties to file: $e", e))
       }
   }
 
-  def addProperties(properties: Properties, dataset: Dataset): Try[Unit] = {
+  def addProperties(properties: Properties, dataset: Dataset, emailaddress: String)(implicit settings: Settings): Try[Unit] = {
```
rvanheest: Maybe you should give the datamanager's id as an argument instead of giving the whole `Settings` object here?

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -15,33 +15,78 @@
...
+    // Note that the datamanager 'precondition' is checked when datamanagerEmailaddress is evaluated the first time
+    if(datamanagerEmailaddress == null) {
+      val id = settings.datamanager
+      datamanagerEmailaddress = settings.ldap.query(id)(a => a)
+        .flatMap(attrsSeq => {
+          if (attrsSeq.isEmpty) Failure(new ActionException(row, s"""DatamanagerID "$id" is unknown"""))
+          else if (attrsSeq.size > 1) Failure(new ActionException(row, s"""There appear to be multiple users with id "$id""""))
+          else Success(attrsSeq.head)
+        })
+        .flatMap(attrs => {
```
rvanheest, line 66: _Note to the careful reviewer:_ these `flatMap`s are actually `filter`s, but with custom error messages!

rvanheest (changes requested):
Great job! Here are some minor points that might need some attention.

src/main/scala/nl/knaw/dans/easy/multideposit/CommandLineOptions.scala
PaulBoon: Ok, username (id) is probably better

rvanheest:
please use this ReadmeSpec instead: https://github.com/rvanheest-DANS-KNAW/easy-module-archetype/blob/9ac2d1581e7932364d8470c60c504f265fec6a73/src/main/resources/archetype-resources/src/test/scala/ReadmeSpec.scala

src/test/scala/nl/knaw/dans/easy/multideposit/ReadmeSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+    Console.withOut(mockedStdOut) {
+      clo.printHelp()
+    }
+    mockedStdOut.toString
+  }
+
+  "arguments" should "be part of README.md" in {
+    val options = helpInfo.replaceAll("\\(default[^)]+\\)","").split("Options:")(1)
+    new File("README.md") should containTrimmed(options)
+  }
```
jo-pol: Pleas replace the one above with 
```
  "options in help info" should "be part of README.md" in {
    val lineSeparators = s"(${System.lineSeparator()})+"
    val options = helpInfo.split(s"${lineSeparators}Options:$lineSeparators")(1)
    options.trim.length shouldNot be (0)
    new File("README.md") should containTrimmed(options)
  }
```

src/test/scala/nl/knaw/dans/easy/multideposit/ReadmeSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+  val mockedProps = {
+    val ps = new PropertiesConfiguration()
+    ps.setDelimiterParsingDisabled(true)
+    ps.load(new File(RES_DIR_STR + "/debug-config", "application.properties"))
+    ps
+  }
+
+  val mockedArgs = Array("-s", RES_DIR_STR, RES_DIR_STR + "/allfields/input", RES_DIR_STR + "/allfields/output", "datamanager")
+
+  val clo = new ScallopCommandLine(mockedProps, mockedArgs)
```
jo-pol, line 35: To get rid of the mocking problems add override verify:

```
  private val clo = new ScallopCommandLine(Array[String]()) {
    // avoids System.exit() in case of invalid arguments or "--help"
    override def verify(): Unit = {}
  }
```
Should also be done for `easy-stage-dataset` Sorry @paulboon for not seeing/remembering it earlier.

jo-pol (changes requested):

PR #46, opened 2017-02-10
EASY-1166 update

lindareijnhoudt:
update on EASY-1166

#### When applied it will
* Override ScallopConf.verify, as suggested by @jo-pol in #45 
* Check whether DATASET, SF_DOMAIN, SF_COLLECTION and SF_USER only contain valid chars
* Log row-number '-1' for preconditions that pertain to the cmdline parameter DATAMANAGER


rvanheest (approved):
LGTM

PR #49, opened 2017-03-08
Easy 1178: create deposit in staging directory, then move it to the output directory

rvanheest:
fixes EASY-1178

#### When applied it will
* add functionality in the debug run script to move the generated data from a previous run to a different folder
* introduce an extra option to the command line interface which sets the staging directory; if not set, the default from the properties file is used
* create the deposit in the staging directory rather than the output directory
* move the deposit to the output directory only when all other actions were successful
* not revert the move if something goes wrong there! (see explanation in that action)

easy-dtap                      | [PR#82](https://github.com/DANS-KNAW/easy-dtap/pull/82)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ ARGUMENTS
                                      'staging-dir' in 'application.properties' is used. (default = data/staging)
         --help                       Show help message
         --version                    Show version of this program
-  
+
    trailing arguments:
     multi-deposit-dir (required)   Directory containing the Submission Information Package to process. This must
                                    be a valid path to a directory containing a file named 'instructions.csv' in
--- a/README.md
+++ b/README.md
@@ -43,22 +43,26 @@ ARGUMENTS
 ```
   Usage: 
 
-      easy-split-multi-deposit.sh [{--springfield-inbox|-s} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
+      easy-split-multi-deposit.sh [{--springfield-inbox|-s} <dir>] [{--staging-dir|-d} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
 
   Options:
 
     -s, --springfield-inbox  <arg>   The inbox directory of a Springfield Streaming Media Platform installation.
-                                     If not specified the value of springfield-inbox in application.properties
-                                     is used. (default = data/springfield-inbox)
+                                     If not specified the value of 'springfield-inbox' in
+                                     'application.properties' is used. (default = data/springfield-inbox)
+    -d, --staging-dir  <arg>         A directory in which the deposit directories are created, after which they
+                                     will be moved to the 'deposit-dir'. If not specified, the value of
+                                     'staging-dir' in 'application.properties' is used. (default = data/staging)
         --help                       Show help message
         --version                    Show version of this program
-
+  
    trailing arguments:
     multi-deposit-dir (required)   Directory containing the Submission Information Package to process. This must
                                    be a valid path to a directory containing a file named 'instructions.csv' in
                                    RFC4180 format.
-    deposit-dir (required)         A directory in which the deposit directories must be created. The deposit
-                                   directory layout is described in the easy-sword2 documentation
+    deposit-dir (required)         A directory to which the deposit directories are moved after the staging has
+                                   been completed successfully. The deposit directory layout is described in the
+                                   easy-sword2 documentation
     datamanager (required)         The username (id) of the datamanger (archivist) performing this deposit
 ```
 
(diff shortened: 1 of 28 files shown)


README.md
```diff
@@ -43,22 +43,26 @@ ARGUMENTS
...
-      easy-split-multi-deposit.sh [{--springfield-inbox|-s} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
+      easy-split-multi-deposit.sh [{--springfield-inbox|-s} <dir>] [{--staging-dir|-d} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
 
   Options:
 
     -s, --springfield-inbox  <arg>   The inbox directory of a Springfield Streaming Media Platform installation.
-                                     If not specified the value of springfield-inbox in application.properties
-                                     is used. (default = data/springfield-inbox)
+                                     If not specified the value of 'springfield-inbox' in
+                                     'application.properties' is used. (default = data/springfield-inbox)
```
jo-pol, line 52: I wonder if backticks would hurt command line help. It could improve the readme.

src/main/scala/nl/knaw/dans/easy/multideposit/actions/MoveDepositToOutputDir.scala
```diff
@@ -0,0 +1,52 @@
...
+    val stagingDirectory = stagingDir(datasetID)
+
+    debug(s"moving $stagingDirectory to $outputDir")
+
+    Try { stagingDirectory.moveDir(outputDir) } recover {
+      case NonFatal(e) => println(s"An error occurred while moving $stagingDirectory to " +
+        s"$outputDir: ${e.getMessage}. This move is NOT revertable! When in doubt, contact your " +
+        s"application manager.", e)
+    }
+  }
```
jo-pol: Can we somehow check if both locations are on the same device? If so "while moving" is not applicable as then a move of a directory is just attaching it to another parent, so its all or nothing.

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
```diff
@@ -145,7 +148,7 @@ package object multideposit {
      *
      * @param data the content to write to the file
      */
-    @throws(classOf[IOException])
+    @throws[IOException]("in case of an I/O error")
```
jo-pol, line 151: this is new to me, ah required when overriding in [java](http://stackoverflow.com/questions/25129848/scalas-throws-annotation-is-ignored-in-javac-if-i-declare-the-variable-as-its/25131126#25131126)

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
```diff
@@ -254,6 +276,10 @@ package object multideposit {
   }
 
   implicit class DatasetExtensions(val dataset: Dataset) extends AnyVal {
+    def getRowNumber: Int = {
+      dataset("ROW").head.toInt // first occurrence of dataset, assuming it is not empty
```
jo-pol, line 280: If `dataset("ROW")` is a stream, you discard the first row with this action. Seems an unexpected side effect to me.

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
```diff
@@ -339,41 +365,48 @@ package object multideposit {
...
+  }
+
   // mdDir/datasetID/
   def multiDepositDir(datasetID: DatasetID)(implicit settings: Settings): File = {
     new File(settings.multidepositDir, datasetID)
   }
   // mdDir/instructions.csv
   def multiDepositInstructionsFile(implicit settings: Settings): File = {
-    new File(settings.multidepositDir, instructionsFileName)
+    multiDepositInstructionsFile(settings.multidepositDir)
```
jo-pol, line 381: I have a vague memory the other should be made final to prevent overriding this one implicitly with the other.

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
rvanheest: What do you mean "_If `dataset("ROW")` is a stream_"? `Dataset` has type `Map[String, List[String]]`, so `dataset("ROW")` has type `List[String]`. Also, `head` takes the first element from a list, if the list is non-empty.

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
rvanheest: no need, we are in a package object here

README.md
rvanheest: they're already in a code block, so no need for that

src/main/scala/nl/knaw/dans/easy/multideposit/actions/MoveDepositToOutputDir.scala
rvanheest: We could ask whether the `outputDir` exists. If so, this means the `outputDir` is on another file system, in which the move action might have failed partially and have partially succeeded, meaning that a partial deposit may be ingested. If the `outputDir` doesn't exist, the `outputDir` is on the same filesystem, in which case nothing has happened, so less panic needs to be put into the error message.

rvanheest: @jo-pol that was a great suggestion. See 52a7a0c.

PR #51, opened 2017-03-16
EASY-1231 Fix swapped XY ccordinates for RD

PaulBoon:
fixes EASY-1231

#### When applied it will
* fix the swapping

createSpatialBoxes and createSpatialPoints in
easy-split-multi-deposit/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala

and 
easy-split-multi-deposit/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
Look for 
` "createDcmiMetadata" should "return the expected spatial Box elements" in`
and
`"createDcmiMetadata" should "return the expected spatial Point elements" in `

### DIFF
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
@@ -218,31 +218,75 @@ object AddDatasetMetadataToDeposit {
   }
 
   def createSpatialPoints(dataset: Dataset): Seq[Elem] = {
-    // coordinate order latitude (DCX_SPATIAL_Y), longitude (DCX_SPATIAL_X)
-    dataset.rowsWithValuesForAllOf(composedSpatialPointFields).map(mdKeyValues =>
+    dataset.rowsWithValuesForAllOf(composedSpatialPointFields).map(mdKeyValues => {
+      val srsName = createSrsName(mdKeyValues)
+      // coordinate order x, y = longitude (DCX_SPATIAL_X), latitude (DCX_SPATIAL_Y)
+      val xy = s"${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")}"
+      // coordinate order y, x = latitude (DCX_SPATIAL_Y), longitude (DCX_SPATIAL_X)
+      val yx = s"${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")}"
+
+      val pos = srsName match {
+        case "http://www.opengis.net/def/crs/EPSG/0/28992" => xy
+        case "http://www.opengis.net/def/crs/EPSG/0/4326" => yx
+        case _ => yx
+      }
+
       // @formatter:off
-      <dcx-gml:spatial srsName={createSrsName(mdKeyValues)}>
+      <dcx-gml:spatial srsName={srsName}>
         <Point xmlns="http://www.opengis.net/gml">
-          <pos>{mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")} {mdKeyValues.getOrElse("DCX_SPATIAL_X", "")}</pos>
+          <pos>{pos}</pos>
         </Point>
       </dcx-gml:spatial>
       // @formatter:on
-    )
+    })
   }
 
+  /*
+   Note that X is along North - South and Y is along East - West
+   The lower corner is with the minimal coordinate values and upper corner with the maximal coordinate values
+   If x was increasing from West to East and y was increasing from South to North we would have
+   lower corner (x,y) = (West,South) and upper corner (x,y) = (East,North)
+   This is the case for the WGS84 and RD coordinate systems, but not per se for any other system!
+
+                         upper(x,y)=(E,N)
+                *---N---u
+                |       |
+                W       E
+                |       |
+    ^           l---S---*
+    |           lower(x,y)=(W,S)
+    y
+     x -->
+
+   */
   def createSpatialBoxes(dataset: Dataset): Seq[Elem] = {
-    dataset.rowsWithValuesForAllOf(composedSpatialBoxFields).map(mdKeyValues =>
+    dataset.rowsWithValuesForAllOf(composedSpatialBoxFields).map(mdKeyValues => {
+      val srsName = createSrsName(mdKeyValues)
+
+      val lowerX = mdKeyValues.getOrElse("DCX_SPATIAL_WEST", "")
+      val upperX = mdKeyValues.getOrElse("DCX_SPATIAL_EAST", "")
+      val lowerY = mdKeyValues.getOrElse("DCX_SPATIAL_SOUTH", "")
+      val upperY = mdKeyValues.getOrElse("DCX_SPATIAL_NORTH", "")
+      val xy = (s"$lowerX $lowerY", s"$upperX $upperY")
+      val yx = (s"$lowerY $lowerX", s"$upperY $upperX")
+
+      val (lower, upper) = srsName match {
+        case "http://www.opengis.net/def/crs/EPSG/0/28992" => xy
+        case "http://www.opengis.net/def/crs/EPSG/0/4326" => yx
+        case _ => yx
+      }
+
       // @formatter:off
       <dcx-gml:spatial>
         <boundedBy xmlns="http://www.opengis.net/gml">
-          <Envelope srsName={createSrsName(mdKeyValues)}>
-            <lowerCorner>{mdKeyValues.getOrElse("DCX_SPATIAL_NORTH", "")} {mdKeyValues.getOrElse("DCX_SPATIAL_EAST", "")}</lowerCorner>
-            <upperCorner>{mdKeyValues.getOrElse("DCX_SPATIAL_SOUTH", "")} {mdKeyValues.getOrElse("DCX_SPATIAL_WEST", "")}</upperCorner>
+          <Envelope srsName={srsName}>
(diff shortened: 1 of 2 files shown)


src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -218,31 +218,75 @@ object AddDatasetMetadataToDeposit {
   }
 
   def createSpatialPoints(dataset: Dataset): Seq[Elem] = {
-    // coordinate order latitude (DCX_SPATIAL_Y), longitude (DCX_SPATIAL_X)
-    dataset.rowsWithValuesForAllOf(composedSpatialPointFields).map(mdKeyValues =>
+    dataset.rowsWithValuesForAllOf(composedSpatialPointFields).map(mdKeyValues => {
+      val srsName = createSrsName(mdKeyValues)
+      // coordinate order x, y = longitude (DCX_SPATIAL_X), latitude (DCX_SPATIAL_Y)
+      val xy = s"${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")}"
```
rvanheest: put `lazy` before both `xy` and `yx` to make sure only one is computed. Or even better, calculate the `DCX_SPATIAL_X` and `DCX_SPATIAL_Y` separately and do the correct concat in the pattern match on `srsName` below

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -218,31 +218,75 @@ object AddDatasetMetadataToDeposit {
...
+      // coordinate order y, x = latitude (DCX_SPATIAL_Y), longitude (DCX_SPATIAL_X)
+      val yx = s"${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")}"
+
+      val pos = srsName match {
+        case "http://www.opengis.net/def/crs/EPSG/0/28992" => xy
+        case "http://www.opengis.net/def/crs/EPSG/0/4326" => yx
```
rvanheest: both this value and ALL OTHER values return `yx`. Why not merge them by deleting this line? I guess you have a special reason for that?

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -218,31 +218,75 @@ object AddDatasetMetadataToDeposit {
...
+
+                         upper(x,y)=(E,N)
+                *---N---u
+                |       |
+                W       E
+                |       |
+    ^           l---S---*
+    |           lower(x,y)=(W,S)
+    y
+     x -->
```
rvanheest, line 264: didn't you say above that `x` is the vertical axis? (line 245)
that's not what I see here...

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -218,31 +218,75 @@ object AddDatasetMetadataToDeposit {
...
         </Point>
       </dcx-gml:spatial>
       // @formatter:on
-    )
+    })
   }
 
+  /*
+   Note that X is along North - South and Y is along East - West
+   The lower corner is with the minimal coordinate values and upper corner with the maximal coordinate values
```
rvanheest: lower LEFT and upper RIGHT, you mean?
these coordinates are a delicate subject, so better be as specific as possible...

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -218,31 +218,75 @@ object AddDatasetMetadataToDeposit {
...
-    dataset.rowsWithValuesForAllOf(composedSpatialBoxFields).map(mdKeyValues =>
+    dataset.rowsWithValuesForAllOf(composedSpatialBoxFields).map(mdKeyValues => {
+      val srsName = createSrsName(mdKeyValues)
+
+      val lowerX = mdKeyValues.getOrElse("DCX_SPATIAL_WEST", "")
+      val upperX = mdKeyValues.getOrElse("DCX_SPATIAL_EAST", "")
+      val lowerY = mdKeyValues.getOrElse("DCX_SPATIAL_SOUTH", "")
+      val upperY = mdKeyValues.getOrElse("DCX_SPATIAL_NORTH", "")
+      val xy = (s"$lowerX $lowerY", s"$upperX $upperY")
+      val yx = (s"$lowerY $lowerX", s"$upperY $upperX")
```
rvanheest: same applies here for all these values. Only compute them when necessary (`lazy`) or merge them in the pattern match...

src/test/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDepositSpec.scala
```diff
@@ -578,30 +578,31 @@ class AddDatasetMetadataToDepositSpec extends UnitSpec with BeforeAndAfterAll {
...
+            </Envelope>
+          </boundedBy>
         </dcx-gml:spatial>
         <dcx-gml:spatial>
           <boundedBy xmlns="http://www.opengis.net/gml">
             <Envelope srsName="http://www.opengis.net/def/crs/EPSG/0/28992">
-              <lowerCorner>1 3</lowerCorner>
-              <upperCorner>2 4</upperCorner>
+              <lowerCorner>10 30</lowerCorner>
+              <upperCorner>20 40</upperCorner>
```
rvanheest, line 605: mmm, is it right that the values for `lowerCorner` and `upperCorner` are swapped in comparison with the one above and the input?

rvanheest (changes requested):

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
PaulBoon: To make it a bit future proof, you can now add other srsName's with explicit xy or yx, and keep the default yx.

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
PaulBoon: Arrrgh! I swapped it, sorry, will fix it

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
PaulBoon: In this example schematic box it is indeed lower left and upper right, but GML uses the label 'lowerCorner' and 'upperCorner' because you can not be more specific without fixing the direction, which depends on the coordinate system.

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
PaulBoon: Ok

src/test/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDepositSpec.scala
PaulBoon: The order of the coordinates is swapped, that is exactly the issue that needed fixing

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -221,9 +221,9 @@ object AddDatasetMetadataToDeposit {
     dataset.rowsWithValuesForAllOf(composedSpatialPointFields).map(mdKeyValues => {
       val srsName = createSrsName(mdKeyValues)
       // coordinate order x, y = longitude (DCX_SPATIAL_X), latitude (DCX_SPATIAL_Y)
-      val xy = s"${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")}"
+      lazy val xy = s"${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")}"
       // coordinate order y, x = latitude (DCX_SPATIAL_Y), longitude (DCX_SPATIAL_X)
-      val yx = s"${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")}"
+      lazy val yx = s"${mdKeyValues.getOrElse("DCX_SPATIAL_Y", "")} ${mdKeyValues.getOrElse("DCX_SPATIAL_X", "")}"
```
rvanheest: after seeing it again, it might be better to calculate the value for `DCX_SPATIAL_X` and `DCX_SPATIAL_Y` separately (_don't repeat yourself_) and then compose the two values in the correct string inside the pattern match below. You did this already below at line 267-272. Note: in this case you don't need the lazy here, since you use `DCX_SPATIAL_X` and `DCX_SPATIAL_Y` in both cases.

rvanheest (changes requested):
one more little thingy...

rvanheest (approved):
LGTM!

PR #52, opened 2017-03-21
Param action

rvanheest:
**prepares EASY-1167 and EASY-1168**

#### When applied it will
* parameterize `Action` such that `execute` returns `Try[T]` rather than `Try[Unit]`. This allows actions to return computed results, such that they can be reused in multiple other actions. As this is required for EASY-1167 and EASY-1168, I thought I would do this first.
* An example of this reusing is the datamanager email that is retrieved from LDAP (**IO is expensive!**) and therefore only needs to be done once per multi-deposit run. Instead of our former complicated way of doing this by making the email a singleton that defaults to `null`, we refactor this into its own `Action` that returns the email `String`. Then the properties action that requires this email, can get it via function application that is already provided in the `Action`'s API.

Start in `Action.scala`, then continue to `RetrieveDatamanagerAction.scala` and `AddPropertiesToDeposit.scala` and finally look at the integration in `Main.scala`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,16 +41,19 @@
         <dependency>
             <groupId>org.scalatest</groupId>
             <artifactId>scalatest_2.11</artifactId>
+            <version>3.0.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.scalacheck</groupId>
             <artifactId>scalacheck_2.11</artifactId>
+            <version>1.13.2</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.scalamock</groupId>
             <artifactId>scalamock-scalatest-support_2.11</artifactId>
+            <version>3.3.0</version>
             <scope>test</scope>
         </dependency>
 
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
@@ -15,8 +15,6 @@
  */
 package nl.knaw.dans.easy.multideposit
 
-import nl.knaw.dans.easy.multideposit.Action.CombinedAction
-
 import scala.util.{ Failure, Success, Try }
 import nl.knaw.dans.lib.error.{ CompositeException, TraversableTryExtensions }
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
@@ -25,11 +23,9 @@ import scala.annotation.tailrec
 import scala.util.control.NonFatal
 
 /*
-  To future developers: this class is an Applicative Functor. It satisfies the law as defined in
-    - https://wiki.haskell.org/Typeclassopedia#Laws
-    - https://wiki.haskell.org/Typeclassopedia#Laws_2.
+  To future developers: this class is a Category. It should satisfy the law of composition.
  */
-trait Action[+T] extends DebugEnhancedLogging {
+trait Action[-A, +T] extends DebugEnhancedLogging { self =>
 
   private def logPreconditions(): Unit = {
     logger.info(s"Checking preconditions of ${getClass.getSimpleName} ...")
@@ -41,7 +37,7 @@ trait Action[+T] extends DebugEnhancedLogging {
     logger.info(s"An error occurred. Rolling back action ${getClass.getSimpleName} ...")
   }
   protected[Action] def innerCheckPreconditions: Try[Unit] = Try(logPreconditions()).flatMap(_ => checkPreconditions)
-  protected[Action] def innerExecute(): Try[T] = Try(logExecute()).flatMap(_ => execute())
+  protected[Action] def innerExecute(a: A): Try[T] = Try(logExecute()).flatMap(_ => execute(a))
   protected[Action] def innerRollback(): Try[Unit] = Try(logRollback()).flatMap(_ => rollback())
 
   /**
@@ -52,11 +48,12 @@ trait Action[+T] extends DebugEnhancedLogging {
   protected def checkPreconditions: Try[Unit] = Success(())
 
   /**
-   * Exectue the action.
+   * Exectue the action given an input `a`.
    *
+   * @param a the action's input
    * @return `Success` if the execution was successful, `Failure` otherwise
    */
-  protected def execute(): Try[T]
+  protected def execute(a: A): Try[T]
 
   /**
    * Cleans up results of a previous call to run so that a new call to run will not fail because of those results.
@@ -67,13 +64,14 @@ trait Action[+T] extends DebugEnhancedLogging {
 
   /**
    * Run an action. First the precondition is checked. If it fails a `PreconditionsFailedException`
-   * with a report is returned. Else, if the precondition succeed, the action is executed.
+   * with a report is returned. Else, if the precondition succeed, the action is executed given the input `a`.
    * If this fails, the action is rolled back and a `ActionRunFailedException` with a report is returned.
    * If the execution was successful, `Success` is returned
    *
+   * @param a the action's input
    * @return `Success` if the full execution was successful, `Failure` otherwise
    */
-  def run(): Try[T] = {
+  def run(a: A): Try[T] = {
     def reportFailure(t: Throwable): Try[T] = {
       Failure(ActionRunFailedException(
         report = generateReport(
@@ -94,7 +92,7 @@ trait Action[+T] extends DebugEnhancedLogging {
               footer = "Due to these errors in the preconditions, nothing was done."),
             cause = e))
       }
-      t <- innerExecute().recoverWith {
+      t <- innerExecute(a).recoverWith {
         case e1@CompositeException(es) =>
           rollback() match {
             case Success(_) => reportFailure(e1)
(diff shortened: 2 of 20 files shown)


pom.xml
```diff
@@ -41,16 +41,19 @@
         <dependency>
             <groupId>org.scalatest</groupId>
             <artifactId>scalatest_2.11</artifactId>
+            <version>3.0.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.scalacheck</groupId>
             <artifactId>scalacheck_2.11</artifactId>
+            <version>1.13.2</version>
```
rvanheest, line 50: these version increases allow us to use the newest features of these libraries. ScalaCheck, for example, allows for better `Try` support. ScalaTest gives us new operators to verify the contents of sequences and ScalaMock-ScalaTest-Support needed to be updated accordingly to keep working properly.

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -15,22 +15,21 @@
...
- */
 /*
-  To future developers: this class is a semigroup. It satisfies the associativity law as defined
-  in https://wiki.haskell.org/Typeclassopedia#Laws_4.
+  To future developers: this class is an Applicative Functor. It satisfies the law as defined in
+    - https://wiki.haskell.org/Typeclassopedia#Laws
+    - https://wiki.haskell.org/Typeclassopedia#Laws_2.
  */
-trait Action extends DebugEnhancedLogging {
+trait Action[+T] extends DebugEnhancedLogging {
```
rvanheest: This is the major change in this PR. We parameterize `Action` such that it's `run` and `execute` methods return a `Try[T]` rather than `Try[Unit]`. Note that for most subclasses of `Action` we keep `T =:= Unit`, as is shown in file diffs below.

Because of this extra parameter, we have to make a couple of changes in the report functions to make sure the compiler remains happy. Overall the report's output remained the same as before this PR.

The former `combine :: Action -> Action -> Action` combinator has been replaced with a `combineWith :: Action[T] -> Action[S] -> ((T, S) -> R) -> Action[R]` which combines two actions by providing a combinator function `(T, S) -> R`, which combines the outputs of both actions as the final result of the resulting `Action[R]`.

Furthermore we introduced convenience operators `map`, `applyLeft`, `applyRight`, `andThen` and `thenAnd`. See the scaladoc of these operators for more explanation. Notice that `x.andThen(y)` (with `x` and `y` of type `Action[Unit]`) is the same as the former `x.combine(y)`.

src/main/scala/nl/knaw/dans/easy/multideposit/Main.scala
```diff
@@ -40,37 +40,34 @@ object Main extends DebugEnhancedLogging {
 
   def run(implicit settings: Settings): Try[Unit] = {
     MultiDepositParser.parse(multiDepositInstructionsFile)
-      .flatMap(getActions(_).reduce(_ compose _).run)
+      .flatMap(getActions(_).reduce(_ andThen _).run())
   }
 
-  def getActions(datasets: Datasets)(implicit settings: Settings): ListBuffer[Action] = {
+  def getActions(datasets: Datasets)(implicit settings: Settings): ListBuffer[Action[Unit]] = {
```
rvanheest: An example of why we need this parameterized `Action` can already be found in the current code base. The `datamanager` that is provided in the command line arguments needs to be validated and it's email address needs to be queried from LDAP. Given that this is the same for all datasets in this multi-deposit, we only want to execute this query once. Currently we use a very Java-like way with a private (static) variable which is initialized to `null` and allocated with the real emailaddress as the query gets executed. This gives interesting challenges while testing (including private method invocation via reflection) and isn't really Scala-like.

The new implementation makes retrieving the datamanager's email into an atomic `Action`, which is only executed once in this `Main.scala`. It's result (the email address) is injected into the `AddPropertiesToDeposit` action by making the latter of type `Action[String => Unit]` and using the new `applyRight` operator.

rvanheest (commented):
extra info for reviewers

rvanheest:
OK, I confess it was a bit complicated, this change. And hard to understand as well, as I saw yesterday when I tried to explain it to @PaulBoon. Sorry for the headache, Paul!

I tried to improve things by adding documentation and some extra explanation above, but this morning I thought of a simpler way to do the composition that we desire and need.

We parameterize `Action` with an input parameter `A` and an output parameter `T`. This means that the `execute` (to be implemented by the subclasses), `innerExecute` and `run` now take an input of type `A` and return something of type `T`. (_sidenote_, given that `A` is the input parameter, it is in contravariant position (therefore the `-A`) and given that `T` is the output parameter, it is in covariant position (therefore the `+T`))

The new `combine` operator sequentially composes an `Action[A, T]` with another `Action[T, S]`, such that the resulting `Action[A, S]` takes the output of the first `Action` as the input for the second. It therefore should satisfy the composition law `Action(f) combine Action(g) === Action(f compose g)`.

This way I hope it is much less confusing than having just one parameter, which might return a function that needs to be applied separately.

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -68,13 +64,24 @@ trait Action extends DebugEnhancedLogging {
...
    *
+   * @param a the action's input
    * @return `Success` if the full execution was successful, `Failure` otherwise
    */
-  final def run: Try[Unit] = {
+  def run(a: A): Try[T] = {
+    def reportFailure(t: Throwable): Try[T] = {
+      Failure(ActionRunFailedException(
+        report = generateReport(
+          header = "Errors in Multi-Deposit Instructions file:",
```
jo-pol, line 78: The wording made me expect the function to be called on failing preconditions. Shouldn't it be phrased more like 'error during processing ....'?

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -85,101 +92,92 @@ trait Action extends DebugEnhancedLogging {
...
+          }
+        case NonFatal(e1) =>
+          rollback() match {
+            case Success(_) => reportFailure(e1)
+            case Failure(CompositeException(es2)) => reportFailure(CompositeException(List(e1) ++ es2))
+            case Failure(e2) => reportFailure(CompositeException(List(e1, e2)))
+          }
       }
-    } yield ()
+    } yield t
```
jo-pol, line 109: extract methods (into run to limit the number of arguments and of course more logical names) to get something better readable like
```scala

    for {
      _ <- innerCheckPreconditions.recoverWith {
        case NonFatal(e) => recover1(e) 
      }
      t <- innerExecute(a).recoverWith {
        case e1 @ CompositeException(es) => recover2(e1, es)
        case NonFatal(e1) => recover3(e1)
      }
    } yield t
```
2 and 3 are also very much alike...

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -85,101 +92,92 @@ trait Action extends DebugEnhancedLogging {
...
+    header.toOption.fold("")(_ + "\n") +
+      report(List(throwable)).reverse.mkString("\n") +
+      footer.toOption.fold("")("\n" + _)
   }
 
   /**
-   * Compose this action with `other` into a new action such that it runs both actions in the correct order.
-   * [[checkPreconditions]] and [[execute]] are run in order, [[rollback]] is run in reversed order.
+   * Sequentially composes two `Action`s by running their `execute` methods one after the other.
+   * In this, the input of `other` is the output of this `Action`.
```
jo-pol, line 131: Please more details about the execution order:
* preconditions of this
* preconditions of other
* execute of this, on failure rollback of this
* execute of other, on failure rollback of other, then rollback of this

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
```diff
@@ -85,101 +92,92 @@ trait Action extends DebugEnhancedLogging {
...
-  }
+  def combine[S](other: Action[T, S]): Action[A, S] = new Action[A, S] {
+    private var pastSelf = false
 
-  final case class ComposedAction(actions: List[Action]) extends Action {
-    private lazy val executedActions = mutable.Stack[Action]()
+    override def run(a: A): Try[S] = {
+      pastSelf = false
+      super.run(a)
+    }
```
jo-pol, line 143: I don't see how this takes care of the desired execution other, so perhaps more explanation here too.

src/main/scala/nl/knaw/dans/easy/multideposit/actions/RetrieveDatamanagerAction.scala
```diff
@@ -0,0 +1,65 @@
...
+          .filter(_.contains("ARCHIVIST"))
+          .map(_ => Success(attrs))
+          .getOrElse(Failure(ActionException(row, s"""The datamanager "$datamanagerId" is not an archivist""")))
+      })
+      .flatMap(attrs => {
+        Option(attrs.get("mail"))
+          .filter(_.get().toString.nonEmpty)
+          .map(att => Success(att.get().toString))
+          .getOrElse(Failure(ActionException(row, s"""The datamanager "$datamanagerId" does not have an email address""")))
+      })
```
jo-pol, line 63: please extract methods, I didn't yet give the names much thought:
```scala
      .flatMap(attrsSeq => xxx(row, datamanagerId, attrsSeq))
      .flatMap(attrs => attrsOfActive(row, datamanagerId, attrs))
      .flatMap(attrs => attrsOfArchivist(attrs))
      .flatMap(attrs => atrsWithMail(row, datamanagerId, attrs))
```

src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -114,7 +67,7 @@ object AddPropertiesToDeposit {
       .getOrElse(Failure(ActionException(row, """The column "DEPOSITOR_ID" is not present""")))
   }
 
-  def writeProperties(row: Int, datasetID: DatasetID, dataset: Dataset, emailaddress: String)(implicit settings: Settings): Try[Unit] = {
+  def writeProperties(row: Int, datasetID: DatasetID, dataset: Dataset, emailaddress: DatamanagerEmailaddress)(implicit settings: Settings): Try[Unit] = {
```
jo-pol, line 70: Old, but please also extract method in validateDepositor:
```scala
    dataset.get("DEPOSITOR_ID")
      .map(_.filterNot(_.isBlank).toSet)
      .map(uniqueIDs => yyy(uniqueIDs))
      .map(_.flatMap(id => xxx(id)))
      .getOrElse(Failure(ActionException(row, """The column "DEPOSITOR_ID" is not present""")))
```

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
rvanheest: This one is very useful!!! I need to learn to do this more often! Keep reminding me 😃

src/main/scala/nl/knaw/dans/easy/multideposit/Action.scala
rvanheest: this makes sure the `pastSelf` is set to false. After you run the composed action once, the `pastSelf` is set to `true`. When a new run starts, you need to reset `pastSelf` to `false`.

src/main/scala/nl/knaw/dans/easy/multideposit/actions/RetrieveDatamanagerAction.scala
rvanheest: that looks way better indeed!

PR #63, opened 2017-04-20
NO EASY ISSUE: Invalid dataset test

rvanheest:
~~fixes EASY-~~

#### When applied it will
* adjust the README in the `invalidCSV` test dataset to conform to the new output after #60.
* add integration tests that compare the test dataset expected output to the real output
* **also extra's after having done the demo with the archive group:**
    * sort the error messages in the output by line rather than by dataset
    * print the output directory when the whole procedure is a success
    * proper handling of errors with a `row == -1` value
    * only reporting distinct error messages (no more duplicated messages that the datamanagerId is incorrect!)
    * only check voor unique values in columns that require exactly or at most one value per dataset (so you can just drag down in those columns in Excel)
    * improve logging by only calling info once per dataset per `checkPreconditions`/`execute`/`rollback`

@DANS-KNAW/easy for review


src/main/scala/nl/knaw/dans/easy/multideposit/parser/MultiDepositParser.scala
```diff
@@ -69,34 +69,40 @@ class MultiDepositParser(implicit settings: Settings) extends DebugEnhancedLoggi
...
+          }
+          .mkString("\n") +
+        footer.toOption.fold("")("\n" + _)
+    }
+
+    Failure(ParserFailedException(
+      report = generateReport(
+        header = "CSV failures:",
+        throwable = t,
+        footer = "Due to these errors in the 'instructions.csv' nothing was done."),
```
jo-pol: Doesn't this need a colon:
`Due to these errors in the 'instructions.csv', nothing was done.`

rvanheest: not entirely sure about it. The other two similar messages in Action.scala#80 and 90 have a colon there, but I'm doubting whether they should be there. Anyway, add this one here for consistency sake.

PR #65, opened 2017-04-25
fix failing tests on Eko's machine

rvanheest:
~~fixes EASY-~~

#### When applied it will
* fix tests that failed on @ekoi's machine. The issue is that I forgot that the `groupBy` that is used in the `AV_FILE` selection doesn't preserve order in the resulting `Map`. Therefore the order that is stated in the tests worked for me, but did apparently not work for @ekoi. With this fix I no longer compare the lists themselves, but rather the _elements in the lists_.
* hardcode the `Locale` in order to make sure formatting of the `Bag-Size` in `bag-info.txt` is done correctly.

@DANS-KNAW/easy for review
@ekoi thanks for finding this bug!

### DIFF
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddBagToDeposit.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/actions/AddBagToDeposit.scala
@@ -16,6 +16,7 @@
 package nl.knaw.dans.easy.multideposit.actions
 
 import java.io.{ File, FileInputStream }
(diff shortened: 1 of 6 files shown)


rvanheest:
@lindareijnhoudt @ekoi could you please checkout this PR again, run the tests and report on the results? All tests should now pass for both issues.

```bash
# cd to easy-split-multi-deposit
git checkout master
git pull
git checkout blessed/pr/65
mvn test
```

PR #67, opened 2017-05-02
EASY-1257: DC_LANGUAGE / DC_FORMAT / DC_TYPE uitbreiden met xsi:type in DDM

rvanheest:
fixes EASY-1257

#### When applied it will
* add xsi:type attributes to the DDM for DC_LANGUAGE, DC_FORMAT and DC_TYPE
* format stuff with the new formatter
* update the parent pom to 1.44

@DANS-KNAW/easy for review

easy-ddm                      | [PR#6](https://github.com/DANS-KNAW/easy-ddm/pull/6)


src/test/resources/debug-config/formats.txt
```diff
@@ -0,0 +1,13 @@
...
+application/msword
+text/plain
+text/html
+text/sgml
+text/xml
+image/jpeg
+image/gif
+image/tiff
+video/quicktime
+video/mpeg1
```
jo-pol, line 13: Why not copy this file from assembly to the testdir tests?

rvanheest: Seems like a good idea, but the problem is that I already rely on this file in my tests, so I guess we better keep this copy around

PR #71, opened 2017-05-09
Refactorings

rvanheest:
~~fixes EASY-~~

#### When applied it will
* make the AV_FILE and AV_SUBTITLES relative to the deposit rather than the multideposit (as suggested by @janvanmansum)
* rename (as requested by @lindareijnhoudt)
    * dataset -> deposit
    * datasetId -> depositId
    * depositorId -> depositorUserId

@DANS-KNAW/easy for review


rvanheest:
**edit:** in this PR I initially made it so that AV_FILE and AV_SUBTITLES had to be relative to the deposit rather than the multideposit. By means of backwards compatibility I now allow both options but give a warning in case of the latter.

Also I added a check that the AV_FILE and AV_SUBTITLES are not allowed to denote a path; they should point to a specific file!

PR #72, opened 2017-05-11
update documentation + xml changes

rvanheest:
~~fixes EASY-~~

#### When applied it will
* update the documentation according to the former changes
* xml changes that I learned from @jo-pol 

@DANS-KNAW/easy for review

### DIFF
--- a/src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
+++ b/src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
@@ -26,7 +26,7 @@ import org.apache.commons.lang.StringUtils
 
 import scala.collection.JavaConversions.collectionAsScalaIterable
 import scala.util.{ Failure, Success, Try }
-import scala.xml.{ Elem, PrettyPrinter }
+import scala.xml.{ Elem, PrettyPrinter, XML }
 
 package object multideposit {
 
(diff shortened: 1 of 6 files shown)


src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
```diff
@@ -161,17 +161,15 @@ package object multideposit {
...
      * @param encoding the encoding applied to this xml
      */
     @throws[IOException]("in case of an I/O error")
     def writeXml(elem: Elem, encoding: Charset = encoding): Unit = {
-      val header = s"""<?xml version="1.0" encoding="$encoding"?>\n"""
-      val data = new PrettyPrinter(160, 2).format(elem)
-
-      file.write(header + data, encoding)
+      file.getParentFile.mkdirs()
+      XML.save(file.toString, elem, encoding.toString, xmlDecl = true)
```
janvanmansum, line 172: Neat! 👍

src/test/scala/nl/knaw/dans/easy/multideposit/BlackBoxSpec.scala
```diff
@@ -125,4 +128,13 @@ class BlackBoxSpec extends UnitSpec with BeforeAndAfter with MockFactory {
         )
     }
   }
+
+  def transformer(label: String) = new RuleTransformer(new RewriteRule {
```
janvanmansum: Does it need to be called `transformer`? That doesn't convey what it does. It seems to be deleting all the elements named `"$label"`. Is that correct? If so, maybe `removeElementsByName` or something, would better fit the bill.

janvanmansum (approved):

src/main/scala/nl/knaw/dans/easy/multideposit/package.scala
rvanheest: thank you @jo-pol!

src/test/scala/nl/knaw/dans/easy/multideposit/BlackBoxSpec.scala
rvanheest: good one! fixed it

PR #74, opened 2017-06-19
EASY-1286: it is not allowed to have both audio and video files in one dataset

rvanheest:
I turns out that it is not allowed (by easy-ingest-flow) to have both audio and video in a dataset. @jo-pol discovered this morning that this is not checked by easy-split-multi-deposit and that the example datasets actually do contain both audio and video files.

This PR fixes the example datasets by removing the audio from `dataset1` and adding a `dataset4` with a.o. the audio file from `dataset1`. It also fixes the corresponding tests. Besides that, it add a check for this issue.

@DANS-KNAW/easy for review


jo-pol (commented):
Can't spot anything weird.

PaulBoon:
Is there no corresponding Jira issue or any other documentation explaining why it is not allowed?

PaulBoon (approved):

rvanheest:
1. @lindareijnhoudt is supposed to make an issue for this before she merges it.
2. regarding other documentation: this was already implemented by @jo-pol in easy-ingest-flow. Now the result from easy-split-multi-deposit can again be used as input for easy-ingest-flow.

PR #80, opened 2017-08-02
EASY-1314 SF_PLAY_MODE in deposit.properties

rvanheest:
fixes EASY-1314

#### When applied it will
* parse the PlayMode from the csv
* add the PlayMode to `deposit.properties`
* add tests accordingly
* remove an old forgotten version file
* separate the tests in `BlackBoxSpec`
* fix a bug in xml formatting

@DANS-KNAW/easy for review
@jo-pol if you want to do an end-to-end test, you can use this PR...


src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddBagToDeposit.scala
```diff
@@ -37,12 +37,12 @@ case class AddBagToDeposit(deposit: Deposit)(implicit settings: Settings) extend
   }
 
   override def execute(): Try[Unit] = {
-    createBag(deposit).recoverWith {
+    createBag().recoverWith {
```
rvanheest, line 40: this change is unrelated, but was a remainder of an older refactoring...

src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -76,7 +76,8 @@ case class AddPropertiesToDeposit(deposit: Deposit)(implicit settings: Settings)
       "datamanager.email" -> Some(emailaddress),
       "springfield.domain" -> sf.map(_.domain),
       "springfield.user" -> sf.map(_.user),
-      "springfield.collection" -> sf.map(_.collection)
+      "springfield.collection" -> sf.map(_.collection),
+      "springfield.playmode" -> deposit.audioVideo.playMode.map(_.toString)
```
rvanheest, line 80: @jo-pol please make sure the values you read correspond to my output (`springfield.playmode -> x \in {'menu', 'continuous'}`)

lindareijnhoudt:
also add the column SF_PLAY_MODE to the documentation in multi-deposit-instructions.md, with some explanation.

rvanheest:
@jo-pol could you integrate this PR in your manual testing of ingest-flow, given that this PR contains the step before yours.

1. From this PR, take `src/test/resources/allfields/input` into deasy
2. deploy this PR on deasy
3. run `easy-split-multi-deposit <path to input dir> <path to output dir> <datamanager-name (easyadmin)>`
4. after a successful run, the deposit `ruimtereis01` should contain a `springfield.playmode=menu` and `ruimtereis04` should contain a `springfield.playmode=continuous`, while `ruimtereis02` and `ruimtereis03` should not contain a `springfield.playmode` in `deposit.properties`
5. move all deposits to the easy-ingest-flow inbox
6. test your own part and check that the output (in Springfield) is conform both our PRs

jo-pol:
Ran you tests, all four ingested successfully but got springfield-actions for only two out of the four submits:

```
[vagrant@deasy ~]$ ls -la /*/*/springfield-inbox
total 12
drwxr-xr-x. 1 vagrant vagrant 272 Aug  4 11:23 .
drwxr-xr-x. 1 vagrant vagrant 306 Aug  4 10:36 ..
drwxr-xr-x. 1 vagrant vagrant 102 Aug  4 11:20 easy-dataset-15
-rw-r--r--. 1 vagrant vagrant 714 Aug  4 11:20 easy-dataset-15_springfield-actions.xml
drwxr-xr-x. 1 vagrant vagrant 102 Aug  4 11:23 easy-dataset-18
-rw-r--r--. 1 vagrant vagrant 450 Aug  4 11:23 easy-dataset-18_springfield-actions.xml
-rw-r--r--. 1 vagrant vagrant 365 Jan 19  2017 springfield-actions.xml
drwxr-xr-x. 1 vagrant vagrant 136 Jan 19  2017 videos
[vagrant@deasy ~]$ cat /vagrant/shared/springfield-inbox/springfield-actions.xml
<?xml version="1.0" encoding="UTF-8"?>
<actions>
  <add target="/domain/dans/user/janvanmansum/collection/Jans-test-files/presentation/centaur">
    <video src="videos/centaur.mpg" target="0"/>
  </add>
  <add target="/domain/dans/user/janvanmansum/collection/Jans-test-files/presentation/hubble">
    <video src="videos/hubble.mpg" target="0"/>
  </add>
</actions>
[vagrant@deasy ~]$ cat /vagrant/shared/springfield-inbox/easy-dataset-15_springfield-actions.xml
<?xml version='1.0' encoding='UTF-8'?>
<actions xmlns="springfield-actions">
        <add target="/domain/dans/user/janvanmansum/collection/Jans-test-files/">
            <presentation name="easy-dataset:15">
                <video-playlist play-mode="menu">
                  <video src="data/path/to/a/random/video/hubble.mpg" target="hubble.mpg" title="hubble.mpg"></video><video src="data/reisverslag/centaur.mpg" target="centaur.mpg" title="video about the centaur meteorite"><subtitles src="data/reisverslag/centaur.srt" xml:lang="en"/><subtitles src="data/reisverslag/centaur-nederlands.srt" xml:lang="nl"/></video>
                </video-playlist>
            </presentation>
        </add>
    </actions>
[vagrant@deasy ~]$ cat /vagrant/shared/springfield-inbox/easy-dataset-18_springfield-actions.xml
<?xml version='1.0' encoding='UTF-8'?>
<actions xmlns="springfield-actions">
        <add target="/domain/dans/user/janvanmansum/collection/Jans-test-files/">
            <presentation name="easy-dataset:18">
                <video-playlist play-mode="continuous">
                  <audio src="data/path/to/a/random/sound/chicken.mp3" target="chicken.mp3"/>
                </video-playlist>
            </presentation>
        </add>
    </actions>
[vagrant@deasy ~]$```

**Comment by rvanheest** [comment]
that's right, only two of them contain audio/video
those are dataset 1 and 4

File: src/main/scala/nl.knaw.dans.easy.multideposit/package.scala [hunk 3]
```diff
@@ -114,8 +114,8 @@ package object multideposit {
      */
     @throws[IOException]("in case of an I/O error")
     def writeXml(elem: Elem, encoding: Charset = encoding): Unit = {
-      Files.createDirectories(file.getParent)
-      XML.save(file.toString, elem, encoding.toString, xmlDecl = true)
+      Files.createDirectories(path.getParent)
+      XML.save(path.toString, XML.loadString(new PrettyPrinter(160, 2).format(Utility.trim(elem))), encoding.toString, xmlDecl = true)
```
rvanheest, line 115: this was causing a bug in the formatting. To format the generated xml, you put it through the `PrettyPrinter`, but that returns a `String`, so we have to load it into `XML` again, such that the `XML.save` can call `toString` on it...

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
```diff
@@ -19,30 +19,44 @@ import java.nio.file.{ Files, Path, Paths }
...
 import scala.collection.JavaConverters._
 import scala.util.{ Failure, Success }
-import scala.xml.transform.{ RewriteRule, RuleTransformer }
 import scala.xml.{ Elem, Node, NodeSeq, XML }
+import scala.xml.transform.{ RewriteRule, RuleTransformer }
 
-class BlackBoxSpec extends UnitSpec with BeforeAndAfter with MockFactory with CustomMatchers {
+// Note to developers: this classes uses shared tests as described in
+// http://www.scalatest.org/user_guide/sharing_tests
+class BlackBoxSpec extends UnitSpec with MockFactory with CustomMatchers {
```
rvanheest, line 31: I've splitted this file into smaller tests, such that we can see more clearly which part of the whole test failed.

src/test/scala/nl.knaw.dans.easy.multideposit/actions/AddBagToDepositSpec.scala
```diff
@@ -21,11 +21,11 @@ import java.security.MessageDigest
...
 import nl.knaw.dans.easy.multideposit.model.{ Deposit, DepositId }
 import nl.knaw.dans.easy.multideposit.{ Settings, UnitSpec, _ }
-import org.scalatest.BeforeAndAfter
+import org.scalatest.BeforeAndAfterEach
 
-import scala.util.Success
+import scala.util.{ Failure, Success }
 
-class AddBagToDepositSpec extends UnitSpec with BeforeAndAfter {
+class AddBagToDepositSpec extends UnitSpec with BeforeAndAfterEach {
```
rvanheest: `BeforeAndAfterEach` gives better extensibility and is a better code-style overall.

src/test/scala/nl.knaw.dans.easy.multideposit/UnitSpec.scala
```diff
@@ -22,7 +22,7 @@ import nl.knaw.dans.easy.multideposit.model._
 import org.joda.time.DateTime
 import org.scalatest._
 
-abstract class UnitSpec extends FlatSpec with Matchers with OptionValues with Inside with OneInstancePerTest {
+abstract class UnitSpec extends FlatSpec with Matchers with OptionValues with Inside {
```
rvanheest, line 25: `OneInstancePerTest` gives worse performance and results in errors in the new `BlackBoxSpec`

rvanheest (commented):
Some comments about what I'm trying to accomplish. Basically splitting `BlackBoxSpec` into smaller bits, since it was giving me troubles lately.

jo-pol (commented):
Did not yet check the tests

rvanheest:
You might be interested in those, especially the`BlackBoxSpec`, given the `should behave like` construction. Might be useful for you in easy-ingest-flow.

src/main/scala/nl.knaw.dans.easy.multideposit/parser/AudioVideoParser.scala
```diff
@@ -31,18 +31,21 @@ trait AudioVideoParser {
...
     Try {
-      ((springf: Option[Springfield], acc: Option[FileAccessRights.Value], avFiles: Set[AVFile]) => {
-        (springf, acc, avFiles) match {
-          case (None, _, fs) if fs.nonEmpty => Failure(ParseException(rowNum, "The column " +
+      ((springf: Option[Springfield], acc: Option[FileAccessRights.Value], playMode: Option[PlayMode.Value], avFiles: Set[AVFile]) => {
+        (springf, acc, playMode, avFiles) match {
+          case (None, _, _,  fs) if fs.nonEmpty => Failure(ParseException(rowNum, "The column " +
             "'AV_FILE' contains values, but the columns [SF_COLLECTION, SF_USER] do not"))
-          case (s, a, fs) => Try { AudioVideo(s, a, fs) }
+          case (s @ Some(_), a, None, fs) => Try { AudioVideo(s, a, Option(PlayMode.Continuous), fs) }
```
jo-pol, line 38: Are you sure you need a default? As ingest-flow explicitly supports no play-mode.

src/main/scala/nl.knaw.dans.easy.multideposit/package.scala
jo-pol: It is quite a detour to convert the string back to xml and implicitly to a string again just to add the prologue. Can't it be added with some constant, or would that be too error prone?

jo-pol (commented):
Sorry, somehow almost lost the pending comments made so far. The tests need a fresh head.

src/main/scala/nl.knaw.dans.easy.multideposit/parser/AudioVideoParser.scala
rvanheest: yes, first of all this default was specified in the issue; second, this default only applies if the Springfield fields are provided (a.k.a. there is A/V material in it) and no SF_PLAY_MODE is given.

src/main/scala/nl.knaw.dans.easy.multideposit/package.scala
rvanheest: I've been looking in the API and there is only an `XML.save` with a `String` to point to the file and a `Elem` pointing to the xml to be written. In this (and following) method(s) the header is added and the xml is written to the file. Talking about bad API design 😛.
As you noticed yesterday the formatting was messed up without doing the `PrettyPrinter.format` first, so this is the solution I came up with. I want to stay as close to library calls as possible here, rather than copy-pasting stuff and modifying it for our own purposes.

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
```diff
@@ -19,37 +19,63 @@ import java.nio.file.{ Files, Path, Paths }
...
+    Note to future developers:
+    We're sharing tests in this BlackBoxSpec to prevent as much test duplication as possible.
+    However, a test like `"allfields" should behave like allfieldsSpec()`, combined with a
+    `BeforeAndAfterAll` will always first execute the setup stuff from `allfieldsSpec()` and only
+    then run the `beforeAll` and `afterAll` functions. Other traits like `BeforeAndAfter` and
+    `BeforeAndAfterEach` are isomorphic.
+    By not using these traits but manually defining and calling `beforeAll` we make sure this
+    function is called before the shared tests are set up or ran.
+   */
+  def beforeAll(): Unit = {
```
jo-pol, line 61: Isn't this an answer to this [SO question](https://stackoverflow.com/questions/32434009/scalatest-how-to-combine-beforeandaftereach-with-should-behave-like)?

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
```diff
@@ -66,80 +92,159 @@ class BlackBoxSpec extends UnitSpec with BeforeAndAfter with MockFactory with Cu
...
+          "metadata")
+      }
+
+      it should "check bag-info.txt" in {
+        doNotRunOnTravis()
+
+        val bagInfo = bag.resolve("bag-info.txt")
+        val expBagInfo = expBag.resolve("bag-info.txt")
+
+        // skipping the Bagging-Date which is different every time
```
jo-pol, line 146: In other occasions we set the system time to some constant value.

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
```diff
@@ -66,80 +92,159 @@ class BlackBoxSpec extends UnitSpec with BeforeAndAfter with MockFactory with Cu
...
+
+    "ruimtereis01" should behave like bagContents("ruimtereis01", expectedDataContentRuimtereis01)
+    "ruimtereis02" should behave like bagContents("ruimtereis02", expectedDataContentRuimtereis02)
+    "ruimtereis03" should behave like bagContents("ruimtereis03", expectedDataContentRuimtereis03)
+    "ruimtereis04" should behave like bagContents("ruimtereis04", expectedDataContentRuimtereis04)
   }
 
+  beforeAll()
+
+  "allfields" should behave like allfieldsSpec()
```
jo-pol, line 255: I'm confused. Spec is usually an extension for test _classes_ this is a def, too big according to codacy. It starts with an import like a class. If it ... like a duck ...

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
rvanheest: Might be. The issue is that ScalaTest is doing discovery on the tests before they are actually executed. This means that some setup stuff is executed twice: once in the discovery phase, once during the actual test itself. Now the issue I ran into yesterday with this has to do with the part that is executed during discovery. Apparently that includes stuff that is defined within the functions below, but outside the tests that are in the functions (part of the setup). This part is ran before the functions `beforeAll`, etc. That's why we need the manual method call here. Note that we leverage the fact that tests are ran in sequence in ScalaTest, rather than jUnit, which doesn't have a particular order described to its tests.

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
rvanheest: I understand this as saying that you reset the system time for a little while, during which this test is ran. Correct? That assumes that time is paused while running this tests, right? (wauw, so many philosophical questions could come from that...) But does that work on every machine, even if it is very slow?
Would be interested to see some references to the parts of EASY you're referring to...

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
rvanheest: Not sure what you mean, but maybe that's the bit your confused about 😉  Let me explain: there are two tests I want to perform: one for the `allfields` test resource, which is a correctly filled in `instructions.csv` with a correct file set, and one for the `invalidCSV` test resource, which (as the name suggests) has errors in its `instructions.csv`.

In the `allfields` test I want to make sure that every file from all deposits inside the resulting multi-deposit is equal/equivalent to the expected output. Therefore, for each deposit I need to do a number of tests. What I could have done is just write them all out, while copy-pasting all 10 tests for each of the deposits (currently there are 4, so 40 tests, all duplicated, but I foresee more of them coming in the future) and fix a test in multiple places when I made a mistake. Of course we agree that this is not desirable!

To fix this I put them in a function that can be called using the `should behave like` construct in ScalaTest. (Note that `FlatSpec` doesn't allow for nested tests, so that's why I can't put the outside function inside the test on line 246.) This function once performs the creation of the deposits by running `Main.run` and then runs all 40 tests. To really get duplication away and not run any setup code (such as `Main.run`) twice, I had to wrap the 'real' tests in another function and give that function some arguments to abstract over the variables (which is the deposit's name and the list of expected files/folders) and call this function four times with different arguments. This keeps the 'real' tests simple and descriptive in the `it should "..." in {...}`, makes them easy to read and makes them be able to fail on there own without influencing the other tests.

To recapitulate: the spec runs the `allfields` test, which calls the `allfieldsSpec` function, which calls `bagContents` one time for each of the resulting deposits.

Since the `invalidCSV` tests isn't as complicated and doesn't require nested tests, I just used the regular way of testing.

jo-pol: Let me simplify some aspects of my confusion:

Codacy complains about the large method. I'm afraid committers are too busy and won't look beyond the red mark in the list of pull requests  and thus won't notice it is not about a breaking build but about code style. So I'm afraid codacy might be a too strong enforcement for code style.

I saw an example with traits that doesn't seem to need much overhead at http://www.scalatest.org/user_guide/sharing_tests but I don't yet master all the details.

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
jo-pol: from ebiu/OnlineTest.java but there should be more cases

```java
    /**
     * Causes DateTime.now() to return a predefined value.
     * 
     * @param value
     *        a value used in assertions
     */
    private void mockDateTimeNow(final String value) {
        DateTimeUtils.setCurrentMillisFixed(new DateTime(value).getMillis());
    }
```

src/main/scala/nl.knaw.dans.easy.multideposit/parser/AudioVideoParser.scala
lindareijnhoudt: yes, i want a default here, so it will be obvious when running the SMD and creating the bag and deposit.properties what will happen with the springfield-dissemination. it should not depend on the ingest-flow version whether or not it will be in 'continuous' or 'menu' play-mode.

rvanheest:
Fixed merge conflict as well as some related compiler/test warnings. Review/merge can continue.

PR #85, opened 2017-08-16
EASY-1291: move stage folder creation to vagrant.yml

rvanheest:
fixes EASY-1291

#### When applied it will
* move the staging folder creation to the vagrant.yml file

@DANS-KNAW/easy for review


rvanheest:
@jo-pol the conclusion we came up with yesterday of creating the `rpm/` folder was incorrect. Creating any directories, or doing anything else like that is the responsibility of vagrant/ansible, not of the rpm creator. After all, we might decide to change the specified directory (alter it in the application properties in ansible) and create the new directory there. Now, the old directory is still created while installing the rpm...

The scripts in `rpm/` are just hooks into the rpm creation process and the rpm will be created regardless of whether or not to there are any hooks present. Instead, you should actually use `includes/install_dans_module.yml` in easy-dtap to start off with the rpm. If I'm correct, it will pick up the latest release of the rpm, although in some circumstances this is not the latest BUILD version. I'm still figuring this out and hopefully I'll have a look at this with @janvanmansum. Keep you posted on that...

rvanheest:
@jo-pol what I forgot to say in previous comment is that you should do any directory creation and such in vagrant/ansible as shown in this PR... 😉

PR #92, opened 2017-09-22
EASY-1324:  Throw an error if SF columns are available, but no AV format is specified

lindareijnhoudt:
fixes EASY-1324

#### When applied it will
* Throw an error if SF columns are available, but no AV format is specified
* Add new video/ and audio/ media types
* Rename formats.txt to acceptedMediaTypes.txt
* Add a unittest
* Remove copied values of acceptedMediaTypes in BlackBox and debug-config

I'm not nec happy with the Path for the acceptedMediaTypes.txt in BlackBox

I've removed the copied formats.txt from debug-config
the run-reset-env.sh takes the complete assembly/dist/cfg, so this file is already present


rvanheest:
maybe add some text to `multi-deposit-instructions.md` to inform the reader about this connection between springfield and format?

src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -23,11 +23,35 @@ import org.joda.time.format.ISODateTimeFormat
...
 
 case class AddDatasetMetadataToDeposit(deposit: Deposit)(implicit settings: Settings) extends UnitAction[Unit] {
 
+  /**
+   * Verifies whether all preconditions are met for this specific action.
+   *
+   * @return `Success` when all preconditions are met, `Failure` otherwise
+   */
+  override def checkPreconditions: Try[Unit] = {
+    def checkSpringFieldDepositHasAVformat(): Try[Unit] = {
```
rvanheest: add this method to the object below. If you don't need inner-functions, better don't do them...

src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -23,11 +23,35 @@ import org.joda.time.format.ISODateTimeFormat
...
+
```
rvanheest: solution looks good, but try to use a functional style:

```scala
deposit.audioVideo.springfield
  .flatMap(_ => deposit.metadata.formats
    .find(s => s.startsWith("audio/") || s.startsWith("video/"))
    .map(_ => Failure(...)))
  .getOrElse(Success(())
```

src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddDatasetMetadataToDeposit.scala
```diff
@@ -23,11 +23,35 @@ import org.joda.time.format.ISODateTimeFormat
...
+        case(true, false) => Success(())
+        case(true, true) =>
+          Failure(ActionException(deposit.row,
+            "No audio/video Format found for this column: [DC_FORMAT]\n" +
+              "cause: this column should contain at least one " +
+              "audio/ or video/ value because SF columns are present"))
+      }
+    }
+
+    for {
```
rvanheest: no need for a for-comprehension here. If you only want a `()` at the end, just use `map`. Then this call can be a one-liner.

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
```diff
@@ -30,21 +31,12 @@ import scala.xml.{ Elem, Node, NodeSeq, XML }
...
-                          |image/jpeg
-                          |image/gif
-                          |image/tiff
-                          |video/quicktime
-                          |video/mpeg1
-                        """.stripMargin.split("\n").map(_.trim).toSet
+  private val formatsFile: Path = Paths.get("src/main/assembly/dist/cfg/acceptedMediaTypes.txt").toAbsolutePath
+  private val formats =
+    if (Files.exists(formatsFile)) managed(Source.fromFile(formatsFile.toFile)).acquireAndGet(_.getLines.map(_.trim).toSet)
+    else Set.empty[String]
```
rvanheest: maybe we should error here, since this file is expected to be here at all times!
You can use `fail(...)` for that, it's a failing function from ScalaTest.

src/test/scala/nl.knaw.dans.easy.multideposit/BlackBoxSpec.scala
```diff
@@ -61,6 +53,10 @@ class BlackBoxSpec extends UnitSpec with MockFactory with CustomMatchers {
   def beforeAll(): Unit = {
     Paths.get(getClass.getResource("/allfields/input").toURI).copyDir(allfields)
     Paths.get(getClass.getResource("/invalidCSV/input").toURI).copyDir(invalidCSV)
+
+    it should "contain certain Formats" in {
+      formats.contains("audio/mpeg3") shouldBe true
```
rvanheest: I'm not sure whether this test is actually executed inside the `it should ...` context. Never tested that. I usually would write this test without the `it should ...` ceremony.

src/test/scala/nl.knaw.dans.easy.multideposit/actions/AddDatasetMetadataToDepositSpec.scala
```diff
@@ -60,6 +60,21 @@ class AddDatasetMetadataToDepositSpec extends UnitSpec with CustomMatchers with
...
+  "checkPreconditions" should "fail if the deposit contains SF_* fields, but no AV DC_FORMAT is given" in {
+    val deposit = testDeposit1.copy(
+      depositId = depositId,
+      metadata = Metadata(
+        formats = List("text/plain")
+      )
+    )
+    inside(AddDatasetMetadataToDeposit(deposit).checkPreconditions) {
+      case Failure(ActionException(_, message, _)) =>
+        message should {
```
rvanheest: no need to these `{}` since you only test one thing

rvanheest (changes requested):
Looks good already. Here are some small comments.

rvanheest (approved):
LGTM

PR #99, opened 2017-11-30
EASY-1397: fix ordering bug in tests

rvanheest:
fixes EASY-1397

#### When applied it will
* fix a bug in the tests with the ordering of xml elements in `files.xml`, caused by differences in ordering in the traversal of a filesystem

@janvanmansum can you run these tests multiple times, such that we're sure that the tests now run correctly? If you still get failing tests, please post them (including the output) in the comments below. Thanks!

@DANS-KNAW/easy for review

### DIFF
--- a/src/test/scala/nl.knaw.dans.easy.multideposit/CustomMatchers.scala
+++ b/src/test/scala/nl.knaw.dans.easy.multideposit/CustomMatchers.scala
@@ -60,9 +60,27 @@ trait CustomMatchers {
         {
           left.size == right.size && left.map(l => right.exists(l ==)).forall(true ==)
         },
-        s"$left did not contain the same elements as $right",
+        // TODO showing the diffs should be done better, but I don't know yet how to use the IntelliJ supported solution to get a nice diff
+        // need to look deeper into that.
+        s"$left did not contain the same elements as $right: ${ missingLeft(left, right) }; ${ missingRight(left, right) }",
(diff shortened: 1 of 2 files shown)


src/test/scala/nl.knaw.dans.easy.multideposit/CustomMatchers.scala
```diff
@@ -52,4 +53,17 @@ trait CustomMatchers {
...
+    override def apply(left: Iterable[Node]): MatchResult = {
+      MatchResult(
+        {
+          left.size == right.size && left.map(l => right.exists(l ==)).forall(true ==)
+        },
+        s"$left did not contain the same elements as $right",
+        s"$left did contain the same elements as $right"
+      )
+    }
+  }
```
janvanmansum: Maybe print which elements were only present in left and/or right?

janvanmansum (approved):
Looks OK. Tests succeed again. One suggestion (see comment.)

src/test/scala/nl.knaw.dans.easy.multideposit/CustomMatchers.scala
rvanheest: thanks for the suggestion. I'm not entirely happy with the solution for now, but this is the easiest solution for now. Maybe I'll find the 'official' ScalaTest way to do this correctly later.

PR #100, opened 2017-12-22
EASY-1414: Creation timestamp added

gulcinermis:
fixes EASY-1414

#### When applied it will
* add creation.timestamp to deposit.properties

I was not able to test since I could not ssh into deasy due to the maintenance in easy,
but this is the only place I can add the creationTimestamp. Therefore it should work properly.

easy-                      | [PR#](PRlink)


rvanheest (changes requested):
Looks good, but please also add this change to the tests in the corresponding file(s).

rvanheest (approved):

PR #103, opened 2018-02-12
Refactorings

rvanheest:
These are some refactorings I have done on `easy-split-multi-deposit` over the last couple of months. Until now this module depended on the `Action` class that defined rules for combining and evaluation of the various steps in the process. However, this is basically an alternative method for what we do in many of our modules (such as easy-ingest-flow) with the for-comprehensions. I thought I would make this module more like the others and therefore removed the `Action` completely. Now, the various parts of this module are run from a single for-comprehension, making it more readable for those who already know the structure of our other modules.

At first I also had `easy-split-multi-deposit` run as a service as well. (see some commits in this list) However, after some discussion with @lindareijnhoudt we decided we didn't want that feature for now, so I removed that code again.

@DANS-KNAW/easy for review


README.md
```diff
@@ -9,7 +9,8 @@ Utility to process a Multi-Deposit prior to ingestion into the DANS EASY Archive
 SYNOPSIS
 --------
 
-    easy-split-multi-deposit [{--staging-dir|-s} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
+    easy-split-multi-deposit validate [{--staging-dir|-s} <dir>] <multi-deposit-dir> <output-deposits-dir> <datamanager>
```
janvanmansum: As discussed: remove subcommands, convert `validate` to option.

PR #111, opened 2018-06-12
EASY-1511: Let Split Multi Deposit return the base revision UUIDs of the created bags

rvanheest:
fixes EASY-1511

#### When applied it will
* create a CSV report of all deposits that are created using this bulk tool
* add commandline option `--identifier-info` or `-i` to configure the name/location of the report to be created (default to `<user home>/easy-split-multi-deposit-identifier-info-<current-timestamp>.csv`

@DANS-KNAW/easy for review

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ ARGUMENTS
     -s, --staging-dir  <arg>       A directory in which the deposit directories are created, after which they
                                    will be moved to the 'output-deposit-dir'. If not specified, the value of
                                    'staging-dir' in 'application.properties' is used.
-    -v, --validate                 Only validates the input of a Multi-Deposit ingest
+    -v, --validate                 Validates the input of a Multi-Deposit ingest
     -h, --help                     Show help message
         --version                  Show version of this program
   
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Utility to process a Multi-Deposit prior to ingestion into the DANS EASY Archive
 SYNOPSIS
 --------
 
-    easy-split-multi-deposit [{--staging-dir|-s} <dir>] [{--validate|-v}] <multi-deposit-dir> <output-deposits-dir> <datamanager>
+    easy-split-multi-deposit [{--staging-dir|-s} <dir>] [{--identifier-info | -i} <file>] [{--validate|-v}] <multi-deposit-dir> <output-deposits-dir> <datamanager>
 
 
 DESCRIPTION
@@ -48,12 +48,13 @@ ARGUMENTS
 
   Options:
 
-    -s, --staging-dir  <arg>   A directory in which the deposit directories are created, after which they will
-                               be moved to the 'output-deposit-dir'. If not specified, the value of
-                               'staging-dir' in 'application.properties' is used.
-    -v, --validate             Only validates the input of a Multi-Deposit ingest
-    -h, --help                 Show help message
-        --version              Show version of this program
+    -i, --identifier-info  <arg>   CSV file in which information about the created deposits is reported
+    -s, --staging-dir  <arg>       A directory in which the deposit directories are created, after which they
+                                   will be moved to the 'output-deposit-dir'. If not specified, the value of
+                                   'staging-dir' in 'application.properties' is used.
+    -v, --validate                 Only validates the input of a Multi-Deposit ingest
+    -h, --help                     Show help message
+        --version                  Show version of this program
   
    trailing arguments:
     multi-deposit-dir (required)    Directory containing the Submission Information Package to process. This
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
@@ -38,6 +38,7 @@ class SplitMultiDepositApp(formats: Set[String], ldap: Ldap, permissions: Deposi
   private val fileMetadata = new AddFileMetadataToDeposit()
   private val depositProperties = new AddPropertiesToDeposit()
   private val setPermissions = new SetDepositPermissions(permissions)
+  private val reportDatasets = new ReportDatasets()
   private val moveDeposit = new MoveDepositToOutputDir()
 
   override def close(): Unit = ldap.close()
@@ -71,6 +72,7 @@ class SplitMultiDepositApp(formats: Set[String], ldap: Ldap, permissions: Deposi
         }
       }
       _ = logger.info("deposits were created successfully")
+      _ <- reportDatasets.report(deposits)
       _ <- deposits.mapUntilFailure(d => moveDeposit.moveDepositsToOutputDir(d.depositId))
     } yield ()
   }
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/actions/ReportDatasets.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/actions/ReportDatasets.scala
@@ -34,7 +34,7 @@ class ReportDatasets {
 
   private def csvPrinter(file: File): ManagedResource[CSVPrinter] = {
     file.bufferedWriter(charset = encoding)
-      .flatMap[CSVPrinter, ManagedResource](writer => new ManagedResource(new CSVPrinter(writer, csvFormat)))
+      .flatMap[CSVPrinter, ManagedResource](writer => new ManagedResource(csvFormat.print(writer)))
   }
 
   private def printRecord(deposit: Deposit, printer: CSVPrinter): Unit = {
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/actions/ReportDatasets.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/actions/ReportDatasets.scala
@@ -1,14 +1,51 @@
+/**
+ * Copyright (C) 2016 DANS - Data Archiving and Networked Services (info@dans.knaw.nl)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package nl.knaw.dans.easy.multideposit.actions
 
+import better.files.{ ManagedResource, _ }
 import nl.knaw.dans.easy.multideposit.PathExplorer.OutputPathExplorer
+import nl.knaw.dans.easy.multideposit.actions.ReportDatasets._
+import nl.knaw.dans.easy.multideposit.encoding
 import nl.knaw.dans.easy.multideposit.model.Deposit
+import org.apache.commons.csv.{ CSVFormat, CSVPrinter }
 
 import scala.util.Try
 
 class ReportDatasets {
 
   def report(deposits: Seq[Deposit])(implicit output: OutputPathExplorer): Try[Unit] = Try {
-    // TODO implement
-    println(output.reportFile)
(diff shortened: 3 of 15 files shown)


README.md
```diff
@@ -48,12 +48,13 @@ ARGUMENTS
...
-                               be moved to the 'output-deposit-dir'. If not specified, the value of
-                               'staging-dir' in 'application.properties' is used.
-    -v, --validate             Only validates the input of a Multi-Deposit ingest
-    -h, --help                 Show help message
-        --version              Show version of this program
+    -i, --identifier-info  <arg>   CSV file in which information about the created deposits is reported
+    -s, --staging-dir  <arg>       A directory in which the deposit directories are created, after which they
+                                   will be moved to the 'output-deposit-dir'. If not specified, the value of
+                                   'staging-dir' in 'application.properties' is used.
+    -v, --validate                 Only validates the input of a Multi-Deposit ingest
```
jo-pol: 'Only'?

src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
```diff
@@ -38,6 +38,7 @@ class SplitMultiDepositApp(formats: Set[String], ldap: Ldap, permissions: Deposi
   private val fileMetadata = new AddFileMetadataToDeposit()
   private val depositProperties = new AddPropertiesToDeposit()
   private val setPermissions = new SetDepositPermissions(permissions)
+  private val reportDatasets = new ReportDatasets()
```
jo-pol, line 41: I don't expect verbs as class names.

src/main/scala/nl.knaw.dans.easy.multideposit/actions/ReportDatasets.scala
```diff
@@ -0,0 +1,51 @@
...
+import nl.knaw.dans.easy.multideposit.encoding
+import nl.knaw.dans.easy.multideposit.model.Deposit
+import org.apache.commons.csv.{ CSVFormat, CSVPrinter }
+
+import scala.util.Try
+
+class ReportDatasets {
+
+  def report(deposits: Seq[Deposit])(implicit output: OutputPathExplorer): Try[Unit] = Try {
+    for(printer <- csvPrinter(output.reportFile);
```
jo-pol, line 30: why `;`? Isn't this a for comprehension?

src/main/scala/nl.knaw.dans.easy.multideposit/model/Deposit.scala
```diff
@@ -15,26 +15,37 @@
...
-    Deposit(depositId, row, depositorUserId, profile, baseUUID, metadata, fms, audioVideo.springfield)
+    Deposit(depositId = depositId,
+      row = row,
+      depositorUserId = depositorUserId,
+      profile = profile,
+      baseUUID = baseUUID,
+      metadata = metadata,
+      files = fms,
+      springfield = audioVideo.springfield
+    )
```
jo-pol: I don't see much added value in using named values in this case. Having the same on the left and right for most arguments only confuses scala novices. Perhaps rename `fms` instead (`fileMedadataSeq`?).

src/test/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositAppSpec.scala
```diff
@@ -277,10 +284,30 @@ class SplitMultiDepositAppSpec extends TestSupportFixture with MockFactory with
...
+      reportLines(ruimtereis01) should fullyMatch regex s"^$ruimtereis01,$uuidRegex,d5e8f0fb-c374-86eb-918c-b06dd5ae5e71$$"
+      reportLines(ruimtereis02) should fullyMatch regex s"^$ruimtereis02,($uuidRegex),\\1$$"
+      reportLines(ruimtereis03) should fullyMatch regex s"^$ruimtereis03,($uuidRegex),\\1$$"
+      reportLines(ruimtereis04) should fullyMatch regex s"^$ruimtereis04,$uuidRegex,773dc53a-1cdb-47c4-992a-254a59b98376$$"
+    }
+
+    ruimtereis01 should behave like bagContents(ruimtereis01, expectedDataContentRuimtereis01)
+    ruimtereis02 should behave like bagContents(ruimtereis02, expectedDataContentRuimtereis02)
+    ruimtereis03 should behave like bagContents(ruimtereis03, expectedDataContentRuimtereis03)
+    ruimtereis04 should behave like bagContents(ruimtereis04, expectedDataContentRuimtereis04)
```
jo-pol, line 313: 'behave like' is a new concept for me.

src/test/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositAppSpec.scala
```diff
@@ -277,10 +284,32 @@ class SplitMultiDepositAppSpec extends TestSupportFixture with MockFactory with
...
+
+      val header :: lines = reportFile.lines.toList
+      val reportLines = lines.collect {
+        case s if s startsWith ruimtereis01 => ruimtereis01 -> s
+        case s if s startsWith ruimtereis02 => ruimtereis02 -> s
+        case s if s startsWith ruimtereis03 => ruimtereis03 -> s
+        case s if s startsWith ruimtereis04 => ruimtereis04 -> s
+      }.toMap
+      // taken from https://stackoverflow.com/a/6640851/2389405
+      val uuidRegex = "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
```
jo-pol: I see `-[0-9a-f]{4}` repeated 3 times, so: `"[0-9a-f]{8}(-[0-9a-f]{4}){3}-[0-9a-f]{12}"`.

This only matches lower case, would uppercase be valid too?

It would have been fun when we could [predefine] a hex character class.

[predefine]: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html

src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
rvanheest: I understand what you mean. If you have better suggestions for naming (@lindareijnhoudt also invited 😉), let me know.
For now, my thought was to keep naming in line with the other 'actions'. As you probably know, we previously had all these classes (`AddFileMetadata`, `AddPropertiesToDeposit`, `SetDepositPermissions`, etc.) extend from a class called `Action`, hence the class names describe the action being performed. Lately I removed the `Action` class and made it work with a for-comprehension instead, to make it more conform to the other projects. Still, though, the classes have there 'describe-the-action-you-perform' names.
This is also why I came up with a verb in the class name for `ReportDatasets`.

src/main/scala/nl.knaw.dans.easy.multideposit/actions/ReportDatasets.scala
rvanheest: it won't work without!
Notice that this is not a for comprehension: that's with `{...}` and a `yield`, while this is with `(...)` and without a `yield`.

Both this for loop and the for comprehension are syntactic sugar in Scala. The latter rewrites to a series of `map`, `filter` and `flatMap` calls. The former is just a nested set of `foreach` calls.

https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/eb1f9e279055ffafac7959089f0b0a19fa2f917d/src/main/scala/nl.knaw.dans.easy.multideposit/actions/ReportDatasets.scala#L29-L33

In this instance, `csvPrinter(output.reportFile)` returns a `ManagedResource[CSVPrinter]`, for which the `foreach` (1) opens the resource, (2) uses the resource and (3) closes the resource. On the other hand, `deposits` is a `Seq[Deposit]`, so `foreach` iterates over all elements in the collection and performs an action for each of them.

When writing this without for-loop and with `foreach`, it would look like stated below, which probably immediately shows why it's better readable to use a for-loop.

```scala
def report(deposits: Seq[Deposit])(implicit output: OutputPathExplorer): Try[Unit] = Try {
  csvPrinter(output.reportFile).foreach(printer => deposits.foreach(deposit => printRecord(deposit, printer)))
}
```

Finally, notice that the order of the 'resources' in the for-loop matters a whole lot! If the order was reversed, you would create as many `CSVPrinter`s as there are `Deposit`s in the collection!

src/main/scala/nl.knaw.dans.easy.multideposit/model/Deposit.scala
rvanheest: > I don't see much added value in using named values in this case. Having the same on the left and right for most arguments only confuses scala novices.

I mostly agree, but I was forced to. Since I introduced `bagId: BagId = UUID.randomUUID()` to `Deposit`, I wanted to take leverage of this default argument. Therefore I require the names of all parameters

https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/eb1f9e279055ffafac7959089f0b0a19fa2f917d/src/main/scala/nl.knaw.dans.easy.multideposit/model/Deposit.scala#L43-L51

(btw. I moved the `bagId` from the `AddDepositProperties` class to here, since I needed the `bagId` to be used multiple times during the course of the application)

rvanheest: > Perhaps rename fms instead (fileMedadataSeq?).

That name can be better indeed. I changed it to `fileMetadatas`. (first I wanted to use `fileMetadata`, since that is already plural, but then that would be confusing, since it is also single; therefore I just added an extra `s`)

src/test/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositAppSpec.scala
rvanheest: http://www.scalatest.org/user_guide/sharing_tests

It's basically a way to write your tests once and call them for different circumstances.

As you see in this class, `"xxx" should behave like f` defines a function `f` with the actual tests in it. In this use case we also provide arguments to the function to adjust the behaviour of the function according to the specific test case.

I found it to be quite a useful tool; perhaps you can think of some use cases in your code as well...

src/test/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositAppSpec.scala
rvanheest: > This only matches lower case, would uppercase be valid too?

No, upper-case would not be expected in this case: `UUID.randomUUID()` only provides lower case characters `[a-f]`. As explained on [StackOverflow](https://stackoverflow.com/a/13833488/2389405), there is a [standard](http://www.itu.int/rec/T-REC-X.667/en) that says that a UUID should only use lower-case letters. Then it goes on demonstrating that if you use upper-case in `UUID.fromString`, it will be transformed to lower-case in the end. Hence, I will only match on lower-case here.

rvanheest: > I see `-[0-9a-f]{4}` repeated 3 times, so: `"[0-9a-f]{8}(-[0-9a-f]{4}){3}-[0-9a-f]{12}"`.

👍 that is also what [this answer on StackOverflow](https://stackoverflow.com/a/3999043/2389405) suggested in the same question I linked to in the code.
I replaced the original regex with your variant!

rvanheest: > It would have been fun when we could predefine a hex character class.

Not sure what you mean by that, but if you found something cool, definitely feel free to contribute 😃

src/test/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositAppSpec.scala
jo-pol: Thanks, indeed 👍  a solution to a current problem as I'm developing twin tests, one to check against expected output, one to validate against a scheme. Can't have the two in a single test with two `should`s as as the validation would be ignored when running offline.

rvanheest: nice, then those two tests would be inside a function and you would call that function in the `should behave like ...`
curious to see the results!

PR #112, opened 2018-06-14
EASY-1584: Use UUID as the directory name

rvanheest:
fixes EASY-1584

#### When applied it will
* make sure the output directory is the UUID
* add extra logging for report, move and bagIds

@DANS-KNAW/easy for review

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/actions/MoveDepositToOutputDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/actions/MoveDepositToOutputDir.scala
@@ -16,7 +16,7 @@
 package nl.knaw.dans.easy.multideposit.actions
 
 import nl.knaw.dans.easy.multideposit.PathExplorer.{ OutputPathExplorer, StagingPathExplorer }
-import nl.knaw.dans.easy.multideposit.model.DepositId
+import nl.knaw.dans.easy.multideposit.model.{ BagId, DepositId }
 import nl.knaw.dans.lib.error.CompositeException
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 
@@ -24,9 +24,9 @@ import scala.util.{ Failure, Success, Try }
 
 class MoveDepositToOutputDir extends DebugEnhancedLogging {
 
-  def moveDepositsToOutputDir(depositId: DepositId)(implicit stage: StagingPathExplorer, output: OutputPathExplorer): Try[Unit] = {
+  def moveDepositsToOutputDir(depositId: DepositId, bagId: BagId)(implicit stage: StagingPathExplorer, output: OutputPathExplorer): Try[Unit] = {
     val stagingDirectory = stage.stagingDir(depositId)
-    val outputDir = output.outputDepositDir(depositId)
+    val outputDir = output.outputDepositDir(bagId)
 
     logger.debug(s"moving $stagingDirectory to $outputDir")
 
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/actions/MoveDepositToOutputDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/actions/MoveDepositToOutputDir.scala
@@ -15,6 +15,8 @@
  */
 package nl.knaw.dans.easy.multideposit.actions
 
+import java.nio.file.FileAlreadyExistsException
+
 import nl.knaw.dans.easy.multideposit.PathExplorer.{ OutputPathExplorer, StagingPathExplorer }
 import nl.knaw.dans.easy.multideposit.model.{ BagId, DepositId }
 import nl.knaw.dans.lib.error.CompositeException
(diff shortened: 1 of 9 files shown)


src/main/scala/nl.knaw.dans.easy.multideposit/actions/ValidatePreconditions.scala
```diff
@@ -46,10 +46,10 @@ class ValidatePreconditions(ldap: Ldap) extends DebugEnhancedLogging {
       .getOrElse(Success(()))
   }
 
-  def checkOutputDirectoryExists(depositId: DepositId)(implicit output: OutputPathExplorer): Try[Unit] = {
+  def checkOutputDirectoryExists(bagId: BagId, depositId: DepositId)(implicit output: OutputPathExplorer): Try[Unit] = {
```
jo-pol: does the error message still make sense?

src/test/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositAppSpec.scala
```diff
@@ -122,6 +122,42 @@ class SplitMultiDepositAppSpec extends TestSupportFixture with MockFactory with
...
+
+      val header :: lines = reportFile.lines.toList
+      val reportLines = lines.collect {
+        case s if s startsWith ruimtereis01 => ruimtereis01 -> s
+        case s if s startsWith ruimtereis02 => ruimtereis02 -> s
+        case s if s startsWith ruimtereis03 => ruimtereis03 -> s
+        case s if s startsWith ruimtereis04 => ruimtereis04 -> s
+      }.toMap
+      // taken from https://stackoverflow.com/a/6640851/2389405
+      // and https://github.com/DANS-KNAW/easy-split-multi-deposit/pull/111#discussion_r194733478
```
jo-pol: Shouldn't this comment belong to the regexp?

src/main/scala/nl.knaw.dans.easy.multideposit/actions/ValidatePreconditions.scala
rvanheest: The addition of the bagId is because the name of the 'staged bag' (while building the bag) is different than the one from the 'output bag'. This `bagId` (in fact a `UUID`) is assigned by the tool itself, not by the user of the tool. Therefore the user doesn't even know about the `bagId` and its link with the data he/she just deposited. Thus, to indicate that the deposit failed, we need to show the `depositId` instead, which is chosen by the user!

However, now that I think of this somewhat more, I'm not even sure we still need this check. After all, we need to validate that the output directory does not yet exists, but the name is this output directory is the `bagId` (`UUID`) and is therefore (semi-)unique. The chance that this check fails is very minimal and if it fails it is not an error caused by the user. In fact, maybe we should not have this check here at all, but move it to the place where the `bagId` is decided:

```
IF the generated bagId already exists as the name of a directory in the output directory
THEN we need to generate a new bagId and repeat this check
ELSE return the current bagId
```

@jo-pol @lindareijnhoudt @janvanmansum does this make sense and should we implement it this way?

src/test/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositAppSpec.scala
rvanheest: You're right! 👍
I moved the `uuidRegex` to a different place, for it to be reused later on, but forgot to move the comment(s) with it.
Thanks!

rvanheest: a8ff843

src/main/scala/nl.knaw.dans.easy.multideposit/actions/ValidatePreconditions.scala
jo-pol: If we can't trust a UUID to be unique, we'll need such a check anywhere we generate a UUID. Thinking about that, it makes sense to add such a check everywhere we generate a UUID. For example for the [deposit-api](https://github.com/DANS-KNAW/easy-deposit-api/blob/14d05b6d1f937088c753f90c1401315f116cc6fb/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala#L257-L259)

rvanheest: I see what you mean, but if you [read this](https://stackoverflow.com/a/1155027/2389405) you understand why I'm questioning whether this is necessary.

> the annual risk of a given person being hit by a meteorite is estimated to be one chance in 17 billion, which means the probability is about 0.00000000006 (6 × 10−11), equivalent to the odds of creating a few tens of trillions of UUIDs in a year and having one duplicate. In other words, only after generating 1 billion UUIDs every second for the next 100 years, the probability of creating just one duplicate would be about 50%.

src/main/scala/nl.knaw.dans.easy.multideposit/actions/MoveDepositToOutputDir.scala
```diff
@@ -30,7 +32,15 @@ class MoveDepositToOutputDir extends DebugEnhancedLogging {
...
-    Try { stagingDirectory.moveTo(outputDir); () } recoverWith {
+    Try { stagingDirectory.moveTo(outputDir, overwrite = false); () } recoverWith {
+      case e: FileAlreadyExistsException =>
+        Failure(ActionException(s"Could not move $stagingDirectory to $outputDir. The target " +
+          "directory already exists. Since this is only possible when a UUID (universally unique " +
+          "identifier) is not unique; you have hit the jackpot. The chance of this happening is " +
+          "smaller than you being hit by a meteorite. So rejoice in the moment, because this " +
+          "will be a once-in-a-lifetime experience. When you're done celebrating, just try to " +
+          "deposit this and all remaining deposits (be careful not to deposit the deposits that " +
+          "came before this lucky one, because they went through successfully).", e))
```
jo-pol, line 43: Jokes just clutter what a trouble shooter has to fix.

PR #116, opened 2018-08-27
EASY-1684 Laat easy-split-multi-deposit ffprobe aanroepen om te controleren of A/V-bestanden corrupt zijn

janvanmansum:
Fixes EASY-1684

#### When applied it will
* Add a configuration setting to point to the `ffprobe` executable
* Make ffmpeg a dependency through rpm
* Ensure ffprobe is run on each A/V file and an error is reported if the exit code is non-zero

`FfprobeRunner`

* On `deasy` run the roles `_easy-servers-init` (after checking out the EASY-1684 PR for easy-dtap) and `easy-split-multi-deposit`
* Run `easy-split-multi-deposit` on a multi-deposit with one or more corrupt video files in it.

https://github.com/DANS-KNAW/easy-dtap/pull/233

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/Command.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/Command.scala
@@ -19,7 +19,7 @@ import better.files.File
 import nl.knaw.dans.easy.multideposit.PathExplorer.PathExplorers
 import nl.knaw.dans.easy.multideposit.actions.{ InvalidDatamanagerException, InvalidInputException }
 import nl.knaw.dans.easy.multideposit.parser.{ EmptyInstructionsFileException, ParserFailedException }
-import nl.knaw.dans.lib.error.TryExtensions
+import nl.knaw.dans.lib.error.{ CompositeException, TryExtensions }
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import resource.managed
 
@@ -42,7 +42,8 @@ object Command extends App with DebugEnhancedLogging {
       case ParserFailedException(_, _) |
            EmptyInstructionsFileException(_) |
            InvalidDatamanagerException(_) |
-           InvalidInputException(_, _) => // do nothing
+           InvalidInputException(_, _) |
+           CompositeException(_) => // do nothing
       case e => logger.error(e.getMessage, e)
     }
     .doIfFailure { case NonFatal(e) => println(s"FAILED: ${ e.getMessage }") }
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
@@ -19,6 +19,7 @@ import java.io.ByteArrayOutputStream
 
 import better.files.File
 import nl.knaw.dans.easy.multideposit.actions.FfprobeErrorException
+import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.io.IOUtils
 
 import scala.util.{ Failure, Success, Try }
@@ -58,7 +59,7 @@ trait FfprobeRunner {
   }
 }
 
-object FfprobeRunner {
+object FfprobeRunner extends DebugEnhancedLogging{
 
   def apply(exe: File): FfprobeRunner = new FfprobeRunner {
     require(exe exists, "Cannot create ExeRunner for executable that does not exist")
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
@@ -29,8 +29,6 @@ import scala.language.postfixOps
 /**
  * Simple runner for ffprobe. The constructor will fail if the executable does not exist or the current user
  * has no execute permissions.
- *
- * @param ffprobeExe the ffprobe executable to run.
  */
 trait FfprobeRunner {
   val ffprobeExe: File
@@ -61,9 +59,13 @@ trait FfprobeRunner {
 
 object FfprobeRunner extends DebugEnhancedLogging{
 
+  /**
+   * Constructs an FfprobeRunner
+   *
+   * @param exe absolute path of the ffprobe executable
+   * @return
+   */
   def apply(exe: File): FfprobeRunner = new FfprobeRunner {
-    require(exe exists, "Cannot create ExeRunner for executable that does not exist")
-    require(exe isExecutable, "Program is not executable")
     override val ffprobeExe: File = exe
   }
 }
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
@@ -1,0 +1,45 @@
+package nl.knaw.dans.easy.multideposit
+
+import java.io.ByteArrayOutputStream
+
+import better.files.File
+import nl.knaw.dans.easy.multideposit.actions.FfprobeErrorException
+import org.apache.commons.io.IOUtils
+
+import scala.util.{ Failure, Success, Try }
+import scala.sys.process.{ ProcessIO, _ }
+import scala.language.postfixOps
+
+/**
+ * Simple runner for ffprobe. The constructor will fail if the executable does not exist or the current user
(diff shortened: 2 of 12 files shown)


src/main/scala/nl.knaw.dans.easy.multideposit/package.scala
```diff
@@ -35,6 +35,14 @@ package object multideposit {
 
   case class DepositPermissions(permissions: String, group: String)
 
+  /**
+   * An exception caused by the user in some way, for example by passing in erroneous arguments or an invalid
+   * instructions.csv
+   *
+   * @param msg the error message
+   */
+  class UserException(msg: String) extends Exception(msg)
```
rvanheest: doesn't seem to be used at all. please remove.

src/test/scala/nl.knaw.dans.easy.multideposit/actions/ValidatePreconditionsSpec.scala
```diff
@@ -29,7 +29,8 @@ class ValidatePreconditionsSpec extends TestSupportFixture with BeforeAndAfterEa
 
   private val depositId = "dsId1"
   private val ldapMock: Ldap = mock[Ldap]
-  private val action = new ValidatePreconditions(ldapMock)
+  private val ffprobeMock: FfprobeRunner = mock[FfprobeRunner]
```
rvanheest, line 33: This would hardly qualify as testing this new functionality! Please add tests in this class that would cover all control flows in `checkAudioVideoNotCorrupt`. Of course you can use this mocked instance to simulate the `FfprobeRunner`.

src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
```diff
@@ -117,7 +119,15 @@ object SplitMultiDepositApp {
       permissions = configuration.properties.getString("deposit.permissions.access"),
       group = configuration.properties.getString("deposit.permissions.group")
     )
+    val ffprobe = {
+      val ffProbePath = configuration.properties.getString("audio-video.ffprobe")
```
rvanheest, line 123: Would it make sense to do the thing below? That way you would properly catch a null pointer.

```scala
Option(configuration.properties.getString("audio-video.ffprobe"))
  .map(ffProbePath => {
    val exeFile = File(ffProbePath)
    require(exeFile isRegularFile, s"Ffprobe at $exeFile does not exist or is not a regular file")
    require(exeFile isExecutable, s"Ffprobe at $exeFile is not executable")
    FfprobeRunner(exeFile)
  })
  .getOrElse { throw new IllegalArgumentException("Missing configuration for ffprobe") }
```

rvanheest: Besides, you're throwing an exception here (either explicitly as I proposed above, or implicitly using `require`), which is not documented in the type. Would it really hurt the code if we would change the signature to return `Try[SplitMultiDepositApp]`?

src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
```diff
@@ -26,11 +27,12 @@ import nl.knaw.dans.easy.multideposit.parser.MultiDepositParser
 import nl.knaw.dans.lib.error.{ CompositeException, TraversableTryExtensions }
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 
-import scala.util.control.NonFatal
+import scala.language.postfixOps
```
rvanheest, line 30: where do you need this one?

src/main/scala/nl.knaw.dans.easy.multideposit/Command.scala
```diff
@@ -42,7 +42,8 @@ object Command extends App with DebugEnhancedLogging {
       case ParserFailedException(_, _) |
            EmptyInstructionsFileException(_) |
            InvalidDatamanagerException(_) |
-           InvalidInputException(_, _) => // do nothing
+           InvalidInputException(_, _) |
+           CompositeException(_) => // do nothing
```
rvanheest, line 46: Can I try to convince you one more time to deal with `CompositeException` as soon as possible, rather than catching them at the end of the application's run?

`CompositeException` was designed to capture multiple `Throwable`s while running `collectResults`. You can therefore view it as a container that contains a collection of `Throwable` instances. As we do in various other instances, we want to generate an error report out of this collection, to have a user friendly error message.

When we only do this in the root of the application, we usually have no idea where this `CompositeException` is coming from (in itself it doesn't express any semantics of where the error comes from (compared with other error naming such as `EmptyInstructionsFileException` or `InvalidDatamanagerException`)). **Therefore I would say it is best practice to deal with `CompositeException` as early as possible, preferably even immediately after calling `collectResults`.**

Now to be completely honest with you: I did not do this myself in 

https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/e4fe098df0187f6b5293255f06e9aff121512e08/src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala#L54

So maybe you can fix that for me in this PR, since it touches on it.  Thereby you convert `CompositeException` to another type of exception (let's say `InvalidInputException`), which removes the need for you to add `CompositeException` here.

src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
```diff
@@ -0,0 +1,69 @@
...
+        if (exit == 0) Success(())
+        else Failure(FfprobeErrorException(target, exit, new String(err.toByteArray)))
+    }
+  }
+}
+
+object FfprobeRunner extends DebugEnhancedLogging{
+
+  def apply(exe: File): FfprobeRunner = new FfprobeRunner {
+    require(exe exists, "Cannot create ExeRunner for executable that does not exist")
```
rvanheest: aren't you doing the same checks here as in `SplitMultiDepositApp`? Or are these different checks?

rvanheest (changes requested):
Good start. Here are some general comments and questions. Please also find some more review comments in the code below.

1. can someone of us who hasn't got `ffprobe` installed still do a local test run? Or do we all need to install `ffprobe` locally? In either case: what is the correct value of `audio-video.ffprobe` in `src/test/resources/debug-config/application.properties`?
2. could you add a test with invalid audio and video files (that would trigger `ffprobe`)
3. could you add the installation of `ffprobe` to the local VM as well? Then we can test there as well (as I often use to do). Probably it's quite similar to what you did in `easy-dtap`.

src/main/scala/nl.knaw.dans.easy.multideposit/FfprobeRunner.scala
janvanmansum: 👍 Yep, you're right, this is redundant.

src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
janvanmansum: IntelliJ complained about my dot-less calls to `isRegularFile` and `isExecutable` on line 126-7.

src/main/scala/nl.knaw.dans.easy.multideposit/SplitMultiDepositApp.scala
janvanmansum: I am not sure either way. I'll pass on this one, just to limit the scope of this PR. Maybe we should think about this as part of a refactoring of all the error handling in this project.

src/main/scala/nl.knaw.dans.easy.multideposit/package.scala
janvanmansum: 👍 My half-hearted attempt to refactor the error handling left some traces ;)

src/main/scala/nl.knaw.dans.easy.multideposit/Command.scala
janvanmansum: Let's leave the refactoring of the error handling code for another day.

src/test/scala/nl.knaw.dans.easy.multideposit/actions/ValidatePreconditionsSpec.scala
janvanmansum: Man! I was hoping you didn't notice :)

PR #117, opened 2018-09-05
EASY-1705 Add file_visibility column to instructions.csv

janvanmansum:
Fixes EASY-1705

Adding one file attribute turned out to involve a *lot* more code changes than I had expected. IMO the main problem is the over-use of case-blocks, mainly to give more error messages in one run. I have simplified this a bit in a number of places. As a consequence in some cases the user will get only the first error encountered. After correcting that problem and running the tool again, he will then be presented with the next problem. I think this is a fair price to pay. Adding visibility to the pattern match promised to multiply the number of cases, which was already too big.

To do:

- [x] Add unit tests for non-default visibilities.
- [x] Check that visibility is always at least as open as accessibility.
- [x] See if we can generate more error messages about file attributes without resorting to huge match blocks.
- [x] Solve Codacy comments.
- [x] Remove commented-out code.
- [x] Test on `aeasy`.

#### When applied it will
* Add the FILE_VISIBILITY column, with a default value of ANONYMOUS.
* Simplify the code a bit.

Maybe where I started coding, somewhere around [here](https://github.com/DANS-KNAW/easy-split-multi-deposit/compare/master...janvanmansum:EASY-1705?expand=1#diff-8014af9e3a91bac43a2561d2128c9eb3R63).

1. Create an instructions CSV that uses this column for a file
2. Run `easy-split-multi-deposit`.
3. Check that the value ends up in the corresponding deposit's bag, in `metadata/files.xml` in the element for the corresponding file.


src/main/scala/nl.knaw.dans.easy.multideposit/parser/FileDescriptorParser.scala
```diff
@@ -19,88 +19,60 @@ import better.files.File
...
-  def fileDescriptor(depositId: DepositId)(rowNum: => Int)(row: DepositRow): Option[Try[(File, Option[String], Option[FileAccessRights.Value])]] = {
-    val path = row.find("FILE_PATH").map(findPath(depositId))
+  def fileDescriptor(depositId: DepositId)(rowNum: => Int)(row: DepositRow): Option[Try[(File, Option[String], Option[FileAccessRights.Value], Option[FileAccessRights.Value])]] = {
+    def collectErrors(rs: Option[Try[_]]*): Seq[Throwable] = {
+      rs.collect { case Some(Failure(t)) => ParseException(rowNum, t.getMessage, t) }
+    }
+
+    val path = row.find("FILE_PATH").map(findRegularFile(depositId))
     val title = row.find("FILE_TITLE")
     val accessRights = fileAccessRight(rowNum)(row)
```
lindareijnhoudt: should this be renamed to 'accessibility' and 'fileAccessibility' to be more like the 'visibility' part?

PR #124, opened 2018-11-06
fixed and added to the unit tests that are impacted due to this change

stephanHogenboom:
fixes EASY-1779-1781-adds-identifier.doi.registered

#### When applied it will
* deposit.properties files will be created with property identifier.doi.registered=no

The AddPropertiesToDeposit class

create a new deposit and check if the property is added to the deposit.properties file


src/test/resources/allfields/output/input-ruimtereis01/deposit.properties
```diff
@@ -13,3 +13,4 @@ springfield.user=janvanmansum
 springfield.playmode=menu
 state.description=Deposit is valid and ready for post-submission processing
 state.label=SUBMITTED
+identifier.doi.registered=no
```
rvanheest: Could you add newlines at the end of these files. Otherwise we get weird diffs next time...

Also turn on the setting in IntelliJ to add newlines to the end of files on save.

src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddPropertiesToDeposit.scala
```diff
@@ -58,7 +58,8 @@ class AddPropertiesToDeposit extends DebugEnhancedLogging {
       "springfield.domain" -> sf.map(_.domain),
       "springfield.user" -> sf.map(_.user),
       "springfield.collection" -> sf.map(_.collection),
-      "springfield.playmode" -> sf.map(_.playMode.toString)
+      "springfield.playmode" -> sf.map(_.playMode.toString),
+      "identifier.doi.registered" -> Some("no")
```
rvanheest: Please add a comma after this last item. Next time around the diff will be smaller.

rvanheest (changes requested):
Looks good! Just some very minor requests.

src/test/resources/allfields/output/input-ruimtereis01/deposit.properties
stephanHogenboom: Done

src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddPropertiesToDeposit.scala
stephanHogenboom: will do

rvanheest (approved):

PR #134, opened 2019-05-09
urn:uuid prefix for Is-Version-Of property

rvanheest:
Came across this bug while testing something else... The `Is-Version-Of` UUID in `easy-split-multi-deposit` wasn't rendered with a `urn:uuid` prefix.

@DANS-KNAW/easy for review


janvanmansum:
👍

PR #137, opened 2019-05-27
EASY-2100:   add property bag-store.bag-name

stephanHogenboom:
fixes EASY-2100

#### When applied it will
* add property: bag-store.bag-name


rvanheest:
@stephanHogenboom please also add this to the 'cats' branch.

PR #156, opened 2020-05-14
EASY-2423: easy-split-multi-deposit: A/V-bestandsnamen mogen geen spaties bevatten

vesaakerman:
fixes EASY-2423

#### When applied it will
* Check that `A/V filenames` don’t contain `spaces`

easy-                      | [PR#](PRlink)

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.multideposit/parser/ParserValidation.scala
+++ b/src/main/scala/nl.knaw.dans.easy.multideposit/parser/ParserValidation.scala
@@ -16,6 +16,8 @@
 package nl.knaw.dans.easy.multideposit.parser
 
 import cats.syntax.apply._
+import cats.syntax.traverse._
+import cats.instances.list._
 import nl.knaw.dans.common.lang.dataset.AccessCategory
 import nl.knaw.dans.easy.multideposit.model.{ AVFileMetadata, Deposit }
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
@@ -30,6 +32,7 @@ trait ParserValidation extends DebugEnhancedLogging {
       checkSFColumnsIfDepositContainsAVFiles(deposit),
       checkEitherVideoOrAudio(deposit),
       checkAllAVFilesHaveSameAccessibility(deposit),
+      checkAllAVFileNamesWithoutSpaces(deposit),
     ).tupled.map(_ => ())
     // @formatter:on
   }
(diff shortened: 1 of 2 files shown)


src/main/scala/nl.knaw.dans.easy.multideposit/parser/ParserValidation.scala
```diff
@@ -91,4 +94,18 @@ trait ParserValidation extends DebugEnhancedLogging {
...
   }
+
+  def checkAllAVFileNamesWithoutSpaces(deposit: Deposit): Validated[Unit] = {
+    deposit.files.collect { case fmd: AVFileMetadata => fmd.filepath.name }
+      .toList
+      .traverse(name => noSpaces(deposit.row, name))
+      .map(_ => ())
+  }
+
+  private def noSpaces(row: Int, s: String): Validated[String] = {
```
rvanheest: If you curry `noSpaces`, it looks even better 😉

```suggestion
      .traverse(noSpaces(deposit.row))
      .map(_ => ())
  }

  private def noSpaces(row: Int)(s: String): Validated[String] = {
```

src/main/scala/nl.knaw.dans.easy.multideposit/parser/ParserValidation.scala
```diff
@@ -91,4 +94,18 @@ trait ParserValidation extends DebugEnhancedLogging {
...
+    if (!s.contains(' '))
+      s.toValidated
+    else
+      ParseError(row, s"A/V filename '$s' contains spaces").toInvalid
```
rvanheest: usually it's the negation of the check that goes in the `else` branch

```suggestion
    if (s contains ' ')
      ParseError(row, s"A/V filename '$s' contains spaces").toInvalid
    else
      s.toValidated
```

rvanheest (approved):
@vesaakerman I think you're really getting the hang of it! Nice work! The suggestions below are purely stylistic in nature. But the general structure, especially your use of `cats`, looks great!

src/main/scala/nl.knaw.dans.easy.multideposit/parser/ParserValidation.scala
vesaakerman: good idea!

src/main/scala/nl.knaw.dans.easy.multideposit/parser/ParserValidation.scala
vesaakerman: done

PR #157, opened 2020-07-24
EASY-2390: SMD: Zet dans-doi-action op 'update' als DOI en BASE-REVISION zijn gegeven

vesaakerman:
fixes EASY-2390

#### When applied it will
* set `dans-doi-action` to '`update`' when the deposit has a `base-revision` and a `DANS Doi.`
 

easy-dtap                      | [PR#480](https://github.com/DANS-KNAW/easy-dtap/pull/480)

### DIFF
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -5,4 +5,4 @@ auth.ldap.url=ldap://localhost
 auth.ldap.user=cn=ldapadmin,dc=dans,dc=knaw,dc=nl
 auth.ldap.password=changeme
 audio-video.ffprobe=/usr/bin/ffprobe
-dans.doi=10.17026/
+dans.doi.prefix=10.17026/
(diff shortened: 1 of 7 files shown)


src/main/assembly/dist/cfg/application.properties
```diff
@@ -5,3 +5,4 @@ auth.ldap.url=ldap://localhost
 auth.ldap.user=cn=ldapadmin,dc=dans,dc=knaw,dc=nl
 auth.ldap.password=changeme
 audio-video.ffprobe=/usr/bin/ffprobe
+dans.doi=10.17026/
```
lindareijnhoudt:
```suggestion
dans.doi.prefix=10.17026/
```
this should be propagated throughout

lindareijnhoudt: Remember to add extra application properties also to dtap

lindareijnhoudt (changes requested):

src/main/assembly/dist/cfg/application.properties
vesaakerman: done

vesaakerman: also done

